PR #2, opened 2019-09-24
Change blob timeout to 2 seconds

christianbundy:
I was going to increase this to 10 seconds, but thought that 2 seconds should probably be fine. :upside_down_face:


arj03:
Thanks :)

PR #13, opened 2021-01-02
Ssb db2

arj03:
Almost ready. Depends on https://github.com/arj03/ssb-browser-core/pull/9

### DIFF
--- a/ui/profile.js
+++ b/ui/profile.js
@@ -226,7 +226,7 @@ module.exports = function () {
           }, () => {
             alert("followed!") // FIXME: proper UI
             // wait for db sync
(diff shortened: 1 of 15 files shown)


arj03:
There is a problem with restoring feed where the new key does not propagate all the way to ssb-db2, meaning messages are not properly decoded.

ui/profile.js
```diff
@@ -286,6 +264,7 @@ module.exports = function () {
           SSB.syncFeedFromSequence(this.feedId, 0, this.renderProfile)
         else {
           SSB.syncFeedFromLatest(this.feedId, () => {
+            // FIXME: not working
```
arj03, line 267: FIXME

PR #17, opened 2021-01-10
Add graphic for when there is no avatar specified

KyleMaas:
This adds a graphic to be displayed when a user does not have an avatar specified, rather than just display an image error by not specifying an image at all.


arj03:
This looks good. A nice addition. Thanks Kyle :-)

PR #18, opened 2021-01-10
Bump the shrinkwrap version - happens automatically during build

KyleMaas:
Since this happens automatically during build, I figure I might as well commit this and do a pull request so it stops showing up in git diffs.


arj03:
Yep. Thanks

PR #19, opened 2021-01-10
Easy server starting script

KyleMaas:
Adds an easy server starting script which can be run as a regular (non-root) user.


arj03:
Sweet. I use http-server here locally when testing, so nice to see this as a simple script. Thanks

Issue #16, opened 2021-01-10
Unicorn emoji does not show up on Chromium


KyleMaas:
On Chromium on Gentoo Linux, the unicorn emoji next to Connections does not seem to show up.  Maybe a different emoji could be used, or even a dedicated graphic?

arj03:
Strange, it is showing up fine here on Ubuntu and Google Chrome. I'm open to using another emoji :-)

PR #21, opened 2021-01-10
Change connection icon to a green Unicode fisheye

KyleMaas:
Fix for #16 - changes the connection icon from a unicorn to a green fisheye.


arj03:
Thanks

Issue #22, opened 2021-01-10
Support DHT connections


KyleMaas:
It would be great if connections could be made via DHT like they can with Manyverse.  This way you could direct connect without a room or a pub.

arj03:
Yeah that would be great. Manyverse uses the same ssb conn so should be possible to do what it is doing to make it work. Happy to help review and help if you have questions.

KyleMaas:
Well, I've made an attempt with pull request #23.  It probably needs some more work, because I couldn't get it to actually link up with an instance of Manyverse on my phone.  But I feel like I'm quickly getting out of my depth.

KyleMaas:
This should be pretty doable if this gets merged:

https://github.com/staltz/multiserver-dht/pull/5

arj03:
Sweet!

KyleMaas:
Really struggling with this.  I can get it to look like it works with an old version of hyperswarm-web.  Except it only looks like it works, but it actually doesn't.  If I try to update it, current versions give me the following error when I try to build:

`SyntaxError: 'import' and 'export' may appear only with 'sourceType: module' (587:0) while parsing (path-to-local-checkout-of-multiserver-dht)/node_modules/reconnecting-websocket/dist/reconnecting-websocket-mjs.js`

I think I'm going to put this down for a bit and come back to it later.  Just not making any progress.

KyleMaas:
Mwahahaha!!!  I got it to work.  Now to figure out how...

KyleMaas:
There we go.  Between the commits I pushed to my add-dht-support branch and this pull request:

https://github.com/staltz/multiserver-dht/pull/5

And this change to reconnecting-websocket:

https://github.com/pladaria/reconnecting-websocket/issues/149

It now works.  Now, it's still a little buggy and has only had limited testing so far.  But I was able to successfully connect from one browser to another on the same computer and pass messages between them.

KyleMaas:
Oh, and I forgot this issue isn't filed under ssb-browser-core.  You need the add-dht-support branch of that for it to work as well.

KyleMaas:
@arj03 We now have pull request #125 which works without manual dependency patching, if you want to try this out.

KyleMaas:
This is mostly just waiting on this at this point:

https://gitlab.com/staltz/ssb-dht-invite/-/issues/2

KyleMaas:
@arj03 Is it possible you could poke @staltz about this:

[...]

PR #23, opened 2021-01-12
Add support for DHT connections to the Connections screen

KyleMaas:
This adds support for accepting DHT invites via the Connections screen.  Must be used in conjunction with the changes in ssb-browser-core in the pull request at:

https://github.com/arj03/ssb-browser-core/pull/11

I will note that at this time, it does not seem to actually connect.  I'm unsure what the reason is.  But as far as I can tell, ssb-browser-core is able to accept the connection as working, so I have to assume the problem is in the scheduling.  I could use some help with this aspect of this.


arj03:
Overall I think this looks good. I looked at the browser core changes and they seem fine. It would be easier to test if there was a create invite button that called start & create so that one could test this directly between two browsers. Can you add that @KyleMaas?

KyleMaas:
I can try.

KyleMaas:
That adds the button and invite code screen.  It still doesn't actually connect, and I don't understand why, but the functionality on the UI side appears to be working.

KyleMaas:
I think I've figured out why accepting doesn't work.  multiserver-dht uses discovery-swarm, which internally tries to call net.createServer(), which supposedly has iffy support on the browser client side.  It looks like it might be good to try to port multiserver-dht to hyperswarm instead.  See this issue:

https://github.com/staltz/multiserver-dht/issues/3

KyleMaas:
See #22.  I got this to work and was able to successfully test it.

arj03:
Amazing! This is really exciting. Will wait with looking at this until upstream PRs are applied. Will look at the about & suggest very soon.

KyleMaas:
@arj03 We now have pull request #125 which works without manual dependency patching, if you want to try this out.

KyleMaas:
@staltz merged the changes to multiserver-dht.  So now, I think we're down to just needing this:

https://gitlab.com/staltz/ssb-dht-invite/-/issues/2

Once that's in place, I should be able to push a version of this that works without depending in my own patched forks of things.

KyleMaas:

PR #25, opened 2021-01-12
Add a favicon

KyleMaas:
Adds a favicon, based on the hermies.png file.  Main reason is to get rid of the 404 errors when browsers try to request it.


arj03:
Thanks :)

PR #26, opened 2021-01-12
Inline the built files into a single HTML file for easy distribution

KyleMaas:
Builds a file in the dist directory called index-inlined.html which contains all needed assets to run ssb-browser-demo without any further requests to the server.  Basically consolidates the entire program into one easy-to-download, easy-to-host, easy-to-distribute HTML file.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,8 +35,9 @@
     "uglifyify": "^5.0.2"
   },
   "scripts": {
-    "build": "mkdir -p build && browserify -p esmify --full-paths ui/browser.js > build/bundle-ui.js && node write-dist.js && node node_modules/inline-source-cli/dist/index.js --compress false --attribute '' --root dist/ dist/index.html dist/index-inlined.html && sed -i -e \"s~assets/noavatar.svg~data:image/svg+xml;base64,$(base64 -w 0 assets/noavatar.svg)~\" dist/index-inlined.html",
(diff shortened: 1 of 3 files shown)


arj03:
I guess this is an alternative to the workbox PWA. It's nice that its just a single HTML file. Of course you the browser can't cache as much when you make changes, but maybe that is ok. I'm thinking it should maybe be a separate build command?

KyleMaas:
Something like that?

package.json
```diff
@@ -9,6 +9,7 @@
   },
   "dependencies": {
     "human-time": "0.0.2",
+    "inline-source-cli": "^2.0.0",
```
arj03, line 12: this should be a dev dependency

package.json
```diff
@@ -34,8 +35,9 @@
     "uglifyify": "^5.0.2"
   },
   "scripts": {
-    "build": "mkdir -p build && browserify -p esmify --full-paths ui/browser.js > build/bundle-ui.js && node write-dist.js",
-    "release": "mkdir -p build && browserify -g uglifyify -p esmify ui/browser.js > build/bundle-ui.js && node write-dist.js"
+    "build": "mkdir -p build && browserify -p esmify --full-paths ui/browser.js > build/bundle-ui.js",
```
arj03, line 38: This now is missing  && node write-dist.js

arj03:
Almost :) Added two notes.

KyleMaas:
How's that?

(Thanks for your patience with me.  I'm not used to all this newfangled npm-based build stuff.)

arj03:
Perfect. Thanks. You are doing great. It's really nice to see someone come in and just contribute patches. They don't look like much because they are small, but I already love the new favicon :-)

KyleMaas:
Thanks!  What I lack in capabilities, I try to make up for in enthusiasm.  :)

PR #27, opened 2021-01-12
Fixing other Unicode characters which wouldn't display for me

KyleMaas:
Changes Unicode characters for new messages to the envelope character.

### DIFF
--- a/css/main.css
+++ b/css/main.css
@@ -1,5 +1,5 @@
 html {
-  font-family: -apple-system,BlinkMacSystemFont,Segoe UI,Helvetica,Arial,sans-serif,Apple Color Emoji,Segoe UI Emoji,Segoe UI Symbol;
+  font-family: -apple-system,BlinkMacSystemFont,Segoe UI,Helvetica,Arial,sans-serif,Apple Color Emoji,Segoe UI Emoji,Segoe UI Symbol,NotoColorEmoji;
   line-height: 1.5;
   overflow-wrap: break-word;
 }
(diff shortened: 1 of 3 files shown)


arj03:
Is this because the noto font is not loading for you?

KyleMaas:
Yes.  noto-emoji is a separate package on Gentoo.  I installed it, and now I get the party bugle emoji.  But is that something that's a good thing to count on?  If you'd rather stick with that emoji, maybe it'd be a good idea to embed the Noto emoji font?

KyleMaas:
I guess for that matter, the emoji font is also used for post reactions, so maybe it'd be a good idea to embed the font anyway.

arj03:
It should be included? https://github.com/arj03/ssb-browser-demo/blob/master/css/NotoColorEmoji.ttf

KyleMaas:
Hm.  Well, looks like I've got a bug to fix, then.  Uninstalled system-wide noto-emoji, and I'll see if I can hunt it down.

KyleMaas:
Figured it out.  When installed as a system font, Noto Color Emoji was being pulled in as a fallback.  But the embedded version only worked for things in the "emoji" CSS class.  This latest commit adds NotoColorEmoji (the locally-embedded name) as a fallback font for the entire page, which fixes the problems with emojis for me.

arj03:
The css fix looks good. To be honest I liked the notification icons better before, the envelope is very to see on that background.

KyleMaas:
That should revert that change, leaving only the CSS change.  It does seem to work now with the font uninstalled from the system so it can only use the embedded version.

arj03:
Perfect, thanks :)

PR #30, opened 2021-01-12
Add link to profile for each connection

KyleMaas:
Display connections as profile links so when you connect to a pub, you can click on it and see who it's following and/or follow the pub.


arj03:
Thanks

PR #31, opened 2021-01-13
Fix license specifier in package.json

KyleMaas:
Fixes the error about invalid SPDX expression when running npm install.


arj03:
Excellent!

PR #36, opened 2021-01-14
Fix the 'Get msg' buttons

KyleMaas:
If you load a thread containing messages which are outside of your graph, the thread view will show those messages with little "Get msg" links.  If you clicked these, they would not work.  This fixes that so they do work.

### DIFF
--- a/net.js
+++ b/net.js
@@ -68,6 +68,16 @@ SSB.getThread = function(msgId, cb) {
   })
 }
 
+SSB.connected = function(cb) {
+  if(SSB.net && SSB.net.ooo) {
+    cb();
+    return;
+  }
+
+  // Try again later.
+  setTimeout(1000, function() { SSB.connected(cb); });
+}
+
 SSB.getOOO = function(msgId, cb) {
   SSB.connected((rpc) => {
     SSB.net.ooo.get(msgId, cb)
--- a/ui/ssb-msg.js
+++ b/ui/ssb-msg.js
@@ -83,7 +83,7 @@ Vue.component('ssb-msg', {
       SSB.getOOO(this.msg.key, (err, msgValue) => {
         if (err) return alert("Failed to get msg " + err)
 
-        if (net.db.getIndex('contacts').isBlocking(SSB.net.id, msgValue.author))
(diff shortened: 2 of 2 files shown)


net.js
```diff
@@ -68,6 +68,16 @@ SSB.getThread = function(msgId, cb) {
   })
 }
 
+SSB.connected = function(cb) {
```
arj03, line 71: This change is unrelated right?

KyleMaas:
No, that was needed to get the "Get msg" buttons working, because the ssb-msg template's getOOO calls SSB.getOOO, which tries to call SSB.connected.  That fails because SSB.connected didn't exist, so I had to build an SSB.connected function based on inferring from SSB.getOOO what it was supposed to do.

net.js
KyleMaas: (Looks like my reply may have gone to the wrong place.  Trying again.)

No, that was needed to get the "Get msg" buttons working, because the ssb-msg template's getOOO calls SSB.getOOO, which tries to call SSB.connected. That fails because SSB.connected didn't exist, so I had to build an SSB.connected function based on inferring from SSB.getOOO what it was supposed to do.

arj03: Right, that was a leftover. I'll fix that a little bit differently.

arj03:
Thanks for opening this PR. I fixed it a bit differently so gonna close this.

KyleMaas:
That actually doesn't fix one of the issues I ran into.  I initially tried just getting rid of SSB.connected, like you did here:

https://github.com/arj03/ssb-browser-demo/commit/ca194bba462e094cce1093ba9fa6ffac49800f6a#diff-51b24f5be697ccf3d2198c3c57eb757a95ca3c20553f3edf0160af38a5287b5a

But I found that without that, if you clicked "Get msg" before it had actually connected to something (room, pub, etc.), then SSB.net and/or SSB.net.ooo would not exist, and you'd get an exception.  Rebuilding SSB.connected like that resulted in it not only not trying to use those before they exist, but also made the request asynchronous.  I figured that was how it was originally built to work and seemed like a good way to do it, and it worked very well for me.

PR #40, opened 2021-01-15
Forget connections when clicking Disconnect

### LINKED ISSUE #24: Can't disconnect a connection if there are retries left
If a connection fails and there are retries left on the connection, you can't disconnect the connection so that it stops attempting.  You have to wait until it has exhausted all of its retry attempts (either successfully or by failure) before Disconnect buttons work.  If you click them before then, the connection comes right back when the scheduler retries.

KyleMaas:
Allows for ailing connections to be disconnected.  Turns out they were being disconnected, just not forgotten, so the scheduler would try to reconnect them again.  Fixes #24


arj03:
Makes sense

PR #43, opened 2021-01-16
Add a Channels tab to show available channels

KyleMaas:
Pretty self-explanatory.  Adds a tab to the top of the page called Channels which displays a clickable list of the available channels.


arj03:
Looks ok, gonna merge and test :)

arj03:
Nice. Some minor suggestions:
 - Show the number of posts next to to the channel
 - Fix so that #something is the same as something. I get a few of these ##something
 - Maybe sort it by the number of posts
 - This could have a recent or something like that, that only looks at the latest 1000 post messages or so

Anyway excellent work! Really happy to see you productive with the jitdb api.

PR #44, opened 2021-01-16
Implement Channels suggestions from arj03

KyleMaas:
> Nice. Some minor suggestions:
>
>Show the number of posts next to to the channel
>Fix so that #something is the same as something. I get a few of these ##something
>Maybe sort it by the number of posts
>This could have a recent or something like that, that only looks at the latest 1000 post messages or so
>Anyway excellent work! Really happy to see you productive with the jitdb api.

From #43 by @arj03 

Only difference is that I tried 1000 post messages for "recently popular", and it resulted in a still-massive list, so I limited that down to 500.  Feel free to change it if you'd like, though.

### DIFF
--- a/ui/channel.js
+++ b/ui/channel.js
@@ -1,6 +1,6 @@
 module.exports = function () {
   const pull = require('pull-stream')
-  const { and, channel, isPublic, startFrom, paginate, toCallback } = SSB.dbOperators
+  const { and, or, channel, isPublic, type, descending, startFrom, paginate, toCallback } = SSB.dbOperators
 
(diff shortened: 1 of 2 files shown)


KyleMaas:
Oh, I should mention, too, that this fixes a couple bugs in the Channel pages, too:

* Displays posts from not only "#channel" but also "##channel".
* Displays posts in descending order (previously it was displaying in reverse order from everything else).
* Does not display votes as separate posts.

arj03:
Amazing, thank you :)

PR #47, opened 2021-01-16
Add basic Settings screen

KyleMaas:
This adds a basic settings screen with the following options:

* App/tab title (so it can be set to something other than "SSB browser demo").  This changes the tab title so different instances (different private windows for testing, etc.) can be easier identified.
* Number of hops to replicate.
* Caps key.

This ended up being a more difficult change than I expected.  It took me quite a while to figure out that ssb-browser-core had to be pulled in via require() instead of by its browser.js file being packaged into the included bundle.  This also required a bump to the browserify version because it was pulling in an ancient version of path-browserify which was breaking a whole bunch of stuff.

### DIFF
--- a/browser.html
+++ b/browser.html
@@ -18,14 +18,15 @@
       <img src="hermies.png">
     </div>
     <div id="app">
-      <h1>SSB browser demo</h1>
+      <h1>{{ appTitle }}</h1>
       <p id="navigation">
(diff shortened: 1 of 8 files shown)


arj03:
Great!

For now hops is probably not working as it should. I'll be working on updating ssb-friends to ssb-db2 next week, that should make that better.

PR #48, opened 2021-01-16
Fix dependency error looking for ssb-db2/seekers

KyleMaas:
Apparently a dependency error crept in with the new Channels tab, which I didn't spot because ssb-db2 was already installed in node_modules.  This fixes that.


arj03:
I fixed channels to use channel operator instead :-)

PR #50, opened 2021-01-17
Add filter for only showing only messages from people you're directly following

KyleMaas:
Adds a filter for showing only messages from people you're directly following.  Also makes a few other changes:

* Adds a fieldset around filters.
* Adds labels for them so the text is clickable
* Shows a message if there are no messages to display, like if they've all been filtered out.
* Fixes bug where if no messages were found by a query (like if you filtered out everything in that page), you could not load more messages.
* Pagination was inconsistent.  Description said 50 messages, but queries were for 25.
* Displays a message showing which messages are being pulled, so even if all messages are filtered out, you can still see it's looking further back in the timeline.

### DIFF
--- a/ui/public.js
+++ b/ui/public.js
@@ -18,14 +18,17 @@ module.exports = function (componentsState) {
       <textarea class="messageText" v-if="postMessageVisible" v-model="postText"></textarea>
       <button class="clickButton" id="postMessage" v-on:click="onPost">Post new thread</button>
       <input type="file" class="fileInput" v-if="postMessageVisible" v-on:change="onFileSelect">
-      <h2>Last 50 messages</h2>
+      <h2>Last {{ pageSize }} messages</h2>
       <fieldset><legend>Filters</legend>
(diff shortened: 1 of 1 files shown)


arj03:
Looks good from the code :) Thanks!

arj03:
@KyleMaas was testing this, if you don't follow very many people this could easily return 0 results. It would be better to do this as an author query.

arj03:
Other than that I really like the new filter and very nice you cleaned this up.

KyleMaas:
It would indeed be much better to do it at the database query level, but looking at db2's available operators, I couldn't figure out a way to do it.

You mentioned doing an author query.  Do those support supplying a list like an SQL IN?  Or would you "or" all of the followed peoples' IDs (possibly resulting in hundreds of "or"s)?

arj03:
@KyleMaas yeah exactly or'ing them. I implemented it [here](https://github.com/arj03/ssb-browser-demo/commit/1b7e336ce08a2ff9f359627527a8875dd9388e92)

KyleMaas:
Perfect.  Didn't know you could do that.

PR #56, opened 2021-01-17
Add Refresh button for Public messages

### LINKED ISSUE #55: Unchecking filter checkboxes does not take the filter off
After the latest filter changes, if you:

1. Go to the Public tab.
2. Check a filter box.  Observe that the filter takes effect and shows only matching posts.
3. Uncheck the filter box.

Expected result: Filter is removed and everything goes back to displaying as if the filter was not applied.

Actual result: Filter stays applied until the Public tab is reloaded, either by refreshing the page or navigating away from Public and back again (for example, clicking Private and then Public).

KyleMaas:
Also fixes #55 and makes it so that if the "new public message(s)" indicator in the top menu pops up and you click it while you're already on the Public tab, it refreshes the view rather than just re-rendering the existing set of messages.

### DIFF
--- a/ui/new-public-messages.js
+++ b/ui/new-public-messages.js
@@ -18,7 +18,7 @@ module.exports = function (state) {
       reset() {
         // render public resets the newPublicMessages state
         if (this.$route.path == "/public")
-          this.$route.matched[0].instances.default.renderPublic()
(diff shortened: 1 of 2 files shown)


arj03:
Great!

PR #57, opened 2021-01-17
Implement SSB.connected() using events

### LINKED ISSUE #38: Fix SSB.connected
That actually doesn't fix one of the issues I ran into.  I initially tried just getting rid of SSB.connected, like you did here:

https://github.com/arj03/ssb-browser-demo/commit/ca194bba462e094cce1093ba9fa6ffac49800f6a#diff-51b24f5be697ccf3d2198c3c57eb757a95ca3c20553f3edf0160af38a5287b5a

But I found that without that, if you clicked "Get msg" before it had actually connected to something (room, pub, etc.), then SSB.net and/or SSB.net.ooo would not exist, and you'd get an exception.  Rebuilding SSB.connected like that resulted in it not only not trying to use those before they exist, but also made the request asynchronous.  I figured that was how it was originally built to work and seemed like a good way to do it, and it worked very well for me.

_Originally posted by @KyleMaas in https://github.com/arj03/ssb-browser-demo/issues/36#issuecomment-760894156_

KyleMaas:
Implements the missing SSB.connected() function and adds it back into the path for SSB.getOOO().  This actually seems to work quite well - tested by loading a thread with a message outside of my graph, refreshing the page, and clicking "Get msg" before it connected.  It dutifully waited for the connection and then loaded the message.

Fixes #38.

### DIFF
--- a/net.js
+++ b/net.js
@@ -68,7 +68,7 @@ SSB.getThread = function(msgId, cb) {
   })
 }
 
-SSB.isConnected = false
+SSB.activeConnections = 0
 SSB.callbacksWaitingForConnection = []
 function runConnectedCallbacks() {
   while(SSB.callbacksWaitingForConnection.length > 0) {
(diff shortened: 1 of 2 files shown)


net.js
```diff
@@ -68,6 +68,37 @@ SSB.getThread = function(msgId, cb) {
...
+    runConnectedCallbacks()
+  } else {
+    // Register for the connect event so we can keep track of it.
+    SSB.net.on('rpc:connect', (rpc) => {
+      // Now we're connected.  Run all the callbacks.
+      SSB.isConnected = true
+      runConnectedCallbacks()
+
+      // Register an event handler for disconnects so we know to trigger waiting again.
+      rpc.on('closed', () => SSB.isConnected = false)
```
arj03, line 95: minor thing here, but if you are connected to multiple peers this not really correct. I think ssb.conn should have a way to know if you are still connected?

KyleMaas: So you're saying this is incorrect in the use case of:

1. Callback is registered.  isConnected == false
2. Peer 1 connects.  isConnected == true.  Callback is run once
3. Peer 2 connects.  isConnected == true already
4. Peer 1 disconnects.  isConnected == false.  This is where it's incorrect because Peer 2 is still connected.
5. Callback is registered.  Since isConnected == false, it doesn't get run as waits until another peer connects.

If that's what you're referring to, that makes sense, and reference counting should work.  I just based that on how ui/connected.js works.

arj03: yep :)

KyleMaas:
There.  Reference counting for both SSB.connected() as well as the ui/connected.js indicator.  Additionally, this fixes another use case:

1. Peer 1 connects.
2. SSB.connected() is called to register a callback.  Since isConnected == false at this point, events are registered but callback is not run until another peer is connected.

For cases of one peer, the callback would never be run in this case.  So, this fixes that as well.

arj03:
Glorious!

PR #58, opened 2021-01-17
Disallow net: invites

KyleMaas:
This makes it so if a user tries to use a "net:"-based invite, it displays a message explaining that it won't work and why.  It also adds the connection type next to each connection and places each on on its own line instead of relying on the browser to wrap them where we want.


arj03:
Great, thanks :)

KyleMaas:
Related: https://github.com/ssbc/ssb-server/issues/750

PR #59, opened 2021-01-17
Add new SSB.connectedWithData() function and fix profile buttons

### LINKED ISSUE #39: Cannot read property 'partialReplication' of null
Getting the following exception:

```
vue.min.js:6 TypeError: Cannot read property 'partialReplication' of null
    at Object.SSB.syncFeedFromLatest (bundle-ui.js:40)
    at a.downloadMessages (bundle-ui.js:93041)
    at He (vue.min.js:6)
    at HTMLButtonElement.n (vue.min.js:6)
    at HTMLButtonElement.Yr.o._wrapper (vue.min.js:6)
```

This happens if I:
1. Starting with a new profile, add a connection to the Hermies room (http://hermiesroom.staltz.com/).
2. Follow @6CAxOI3f+LUOVrbAl0IemqiS7ATpQvr9Mdw9LC4+Uv0=.ed25519
3. Click "Download Latest Messages for User"

This seems to happen because in SSB.syncFeedFromLatest(), SSB.getPeer() returns null.

KyleMaas:
This adds a new function SSB.connectedWithData() to register a callback for when not only is a peer connected but it's also of a type which has data (not a room).

With this in place, Profile buttons were changed to enqueue their requests if no connection is available that has data.  This should fix #39.

(And I'm sorry the diff's rather large, but most of it's in the extra indentation on the Profile button handlers so they are registered as callbacks with proper indentation level.)

### DIFF
--- a/net.js
+++ b/net.js
@@ -69,10 +69,19 @@ SSB.getThread = function(msgId, cb) {
 }
 
 SSB.activeConnections = 0
+SSB.activeConnectionsWithData = 0
 SSB.callbacksWaitingForConnection = []
+SSB.callbacksWaitingForConnectionWithData = []
 function runConnectedCallbacks() {
   while(SSB.callbacksWaitingForConnection.length > 0) {
     const cb = SSB.callbacksWaitingForConnection.shift()
+    cb(SSB)
+  }
+}
+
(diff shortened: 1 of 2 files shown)


arj03:
nice

PR #70, opened 2021-01-18
Onboarding improvements

KyleMaas:
Instead of having two hardcoded connection defaults (the pub and room) buried in the Connections screen's code, this moves those defaults to the default preferences file.  Additionally, it gives them friendly names on the Connections screen (so people know what they're connecting to and that the two are different), gives them easy-to-use Connect buttons, and removes them from the "Possible connections" list once you've connected to them.  Changing the add type dropdown on the form no longer resets the box's contents, which means if you paste an invite into the box before remembering to use the dropdown, you don't have to paste it again.

This also adds a new single-use onboarding GUI to make it easier for new users to get started.  It only pops up if the app was started without a key (so, first time it's used) and allows the user to choose a name, type a bio, check checkboxes for which of the default connections to add, and adds a few suggested profiles to follow.

### DIFF
--- a/css/main.css
+++ b/css/main.css
@@ -263,3 +263,14 @@ body, html {
   display: inline-block;
   padding: 0.75em;
 }
+
+.get-started-button
+{
+  background-color: #393;
+  color: #fff;
+}
+
+.get-started-button:hover
+{
+  background-color: #6c6;
+}
--- a/defaultprefs.json
+++ b/defaultprefs.json
@@ -3,8 +3,13 @@
   "replicationHops": 1,
   "publicFilters": "",
   "suggestPeers": [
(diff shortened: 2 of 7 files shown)


arj03:
Overall I think this looks fantastic, I'm a little wary of adding specific feeds as suggested, but I think for now this might be fine. Going to merge and test it :)

arj03:
Overall this looks nice, but on my screen I can't see the button as the bottom (needs a scroll it seems) so not able to test this :-)

KyleMaas:
See #77

arj03:
@KyleMaas I ran this on a larger screen now and it synced. Amazon how smooth this process is. Thank you!

KyleMaas:
Thanks!  Glad you like it.  I still don't like how you have to manually refresh the Public tab when it goes from a state of "no messages yet" to "now we've downloaded messages", but that's an improvement that really, really has to be implemented correctly or not at all.  Pages jumping around while you're looking at them can be a maddening problem.

Issue #66, opened 2021-01-18
Names of Connections


KyleMaas:
It would be nice if the Connections screen displayed display names for each connection (when available) instead of just their ID.

KyleMaas:
Waiting to work on this one until #71 gets figured out so I can use SSB.getProfileAsync() to get room-endpoint connections' profile names.

KyleMaas:
#97 partially does this.  I'd like it to show friendly names for room-endpoint connections which are in Suggested, but so far I have yet to figure out how to get profile pulling to work properly when there's no relationship whatsoever yet with the other profile.

KyleMaas:
With the other commit of #97 and #97 being merged, I'd say this one's in good shape to close.

PR #71, opened 2021-01-19
Add 'Send Message' button to Profile pages

### LINKED ISSUE #61: Private Message button on Profile pages
Would be nice if there was a button on the Profile pages that you could click to switch to the New Private Message screen with the person already filled into the To box.

KyleMaas:
Adds a 'Send Message' button to Profile pages.  Also fixes issue with Private tab where if you refreshed the page while you were on the Private tab, it would not load the options for the Recipients list because it tried to do the query before being connected to a peer.

Fixes #61.

### DIFF
--- a/ui/private.js
+++ b/ui/private.js
@@ -45,22 +45,13 @@ module.exports = function (componentsState) {
         var self = this
         if (this.feedId && this.feedId != '') {
           this.postMessageVisible = true
-          SSB.connectedWithData(() => {
-            let rpc = SSB.getPeer()
-            pull(
-              rpc.partialReplication.getMessagesOfType({id: self.feedId, type: 'contact'}),
-              pull.asyncMap(SSB.db.addOOO),
-              pull.collect((err, msgs) => {
-                const profiles = SSB.db.getIndex('profiles').getProfiles()
-                const profile = profiles[self.feedId]
-                if (self.people.length == 0)
-                  self.people = [{ id: self.feedId, name: (profile.name || self.feedId) }]
-                self.recipients = [{ id: self.feedId, name: (profile.name || self.feedId) }]
-
-                // Done connecting and loading the box, so now we can take down the refreshing indicator
-                document.body.classList.remove('refreshing')
-              })
-            )
+          SSB.getProfileAsync(this.feedId, (err, profile) => {
+            if (self.people.length == 0)
+              self.people = [{ id: self.feedId, name: (profile.name || self.feedId) }]
+            self.recipients = [{ id: self.feedId, name: (profile.name || self.feedId) }]
+    
+            // Done connecting and loading the box, so now we can take down the refreshing indicator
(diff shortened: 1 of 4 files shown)


ui/private.js
```diff
@@ -37,18 +40,42 @@ module.exports = function (componentsState) {
...
-        componentsState.newPrivateMessages = false
-
         document.body.classList.add('refreshing')
 
+        var self = this
+        if (this.feedId && this.feedId != '') {
+          this.postMessageVisible = true
+          SSB.connectedWithData(() => {
+            let rpc = SSB.getPeer()
+            pull(
```
arj03, line 50: this seemst to pull from network every time you go to the private messages for a feed view. You should be able to use the partial plugin to see if you need to do that or not.

KyleMaas: There wasn't previously a feed view for the Private tab - I added it for this.  And since it pops up the message compose screen automatically, for it to work consistently with how it would otherwise normally work the recipients list needs to be populated so more recipients could be added.  I also couldn't figure out a way to get the recipient's name to display when only supplied with their ID - I tried using the method that the Profile page uses, but the profiles list doesn't seem to be populated right away, so that didn't work.  I figured since the list had to be populated anyway, and that seemed to require hitting the network, that hitting the network here wasn't any worse.

That's my reasoning, at any rate.  I'd love to know of a better way to fetch a name when given an ID, because that's something I've struggled with when trying other UI improvement ideas (such as my failed attempt at fixing #66).

arj03: Are you looking for these?

```
        const profiles = SSB.db.getIndex('profiles').getProfiles()
        const profile = profiles[feedId]
```

KyleMaas: That's what I tried.  But if you do a fresh load of the page and you haven't already connected to a peer, "profiles" comes back as an empty array.

arj03: Oh sorry now I see, maybe use [this](https://github.com/arj03/ssb-browser-core/blob/62373cdcb0d8bc8a361d6ede4e2ba16707395c08/indexes/contacts.js#L110) to check if you have any contacts for yourself irst?

KyleMaas:
That worked considerably better and is a much more elegant solution I didn't know about.  As such, I modularized that into a SSB.getProfileAsync() function so it can be reused later.

arj03:
Yes, this is a lot better :) Before I hit merge I want to check why getting the profiles are so slow.

arj03:
So the profiles are slow are going to require some more backend work, so lets just merge this for now.

KyleMaas:
Thanks - there were a few things I was waiting to try to take care of because of how many open branches I had open.  Been trying to keep from causing merge conflicts between them.

arj03:
:)

arj03:
This is pretty nice I must say!

KyleMaas:
Thanks!

PR #73, opened 2021-01-19
Delete messages when blocking a user

### LINKED ISSUE #68: Remove data from blocked profiles
As far as I can tell, if you block someone, their data is not deleted from your database.  This could be dangerous in a case where that user has shared something that you do not want.  It would be nice if posts from blocked users could be actively deleted from the database.

KyleMaas:
Clicking "Block and delete" on a profile now blocks them and deletes their messages.

Fixes #68.


arj03:
Excellent!

PR #75, opened 2021-01-19
Fix error caused by early Markdown rendering

KyleMaas:
When markdown contains a blob, rendering it tries to generate a link, but with no peer the URL is generated as an empty string, which causes an error.  This attempts to detect blobs in the body of messages and waits until we're connected to a peer with data to try to render them.

### DIFF
--- a/ui/ssb-msg.js
+++ b/ui/ssb-msg.js
@@ -161,7 +161,9 @@ Vue.component('ssb-msg', {
     if(self.msg.value.content.text.match(blobRegEx)) {
       // It looks like it contains a blob.  There may be better ways to detect this, but this is a fast one.
       // We'll display a sanitized version of it until it loads.
-      self.body = md.markdown(self.msg.value.content.text.replaceAll(blobRegEx, 'Loading...'))
+      if(!SSB.connectedWithData())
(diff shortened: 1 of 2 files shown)


ui/ssb-msg.js
```diff
@@ -157,6 +155,20 @@ Vue.component('ssb-msg', {
     const profiles = SSB.db.getIndex('profiles').getProfiles()
     this.name = getName(profiles, this.msg.value.author)
 
+    // Render the body, which may need to wait until we're connected to a peer.
+    var self = this
+    const blobRegEx = /!\[.*\]\(&.*\)/g
+    if(self.msg.value.content.text.match(blobRegEx)) {
+      // It looks like it contains a blob.  There may be better ways to detect this, but this is a fast one.
+      // We'll display a sanitized version of it until it loads.
+      self.body = md.markdown(self.msg.value.content.text.replaceAll(blobRegEx, 'Loading...'))
```
arj03, line 164: can we use a isConnected thing here so we don't first add the loading body and then the real body in case we are connected?

KyleMaas:
I just pushing another commit that checks for that before adding the placeholders.

arj03:
Sweet, thank you :)

PR #78, opened 2021-01-19
Add scrollbars for onboarding dialog

### LINKED ISSUE #77: Make onboarding screen scrollable
On some devices it is taller than the screen size.  It needs to be scrollable so you can reach the buttons at the bottom.

KyleMaas:
Position the onboarding dialog absolutely instead of fixed so that properly interacts with the browser's normal document scrollbars without having to do any fancy trickery.  This works because since the onboarding dialog only pops up on the initial load, and there are no messages to display at that point, the onboarding dialog should always be the tallest thing on the page, so the user has no way to scroll past it and thus fixed positioning is unnecessary but using it breaks the scrollbars.

Fixes #77.

### DIFF
--- a/css/public.css
+++ b/css/public.css
@@ -2,3 +2,6 @@
   float: right;
 }
 
+#onboarding-dialog {
+  position: absolute;
+}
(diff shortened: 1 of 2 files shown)


arj03:
Thanks

PR #81, opened 2021-01-19
Add ability to open Channel and post message

### LINKED ISSUE #63: Post message from Channel screen
It would be nice if you could navigate to a Channel screen, add a new post to start a new thread, and have it start off with the same channel tag as what you were browsing so you could easily post new messages on a channel.

KyleMaas:
With this, you can open a Channel view (via Channels or clicking a channel link) and post a new message, and it will automatically be placed in that channel.  Also brings over a few of the other features from the Public tab, such as the ability to load more than 50 messages and the ability to refresh the feed.

Fixes #63

### DIFF
--- a/ui/channel.js
+++ b/ui/channel.js
@@ -1,38 +1,102 @@
 module.exports = function () {
   const pull = require('pull-stream')
+  const ssbMentions = require('ssb-mentions')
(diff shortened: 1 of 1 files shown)


KyleMaas:
Oh, and this fixes #63.

arj03:
Looks like an easy merge. Thanks :)

PR #82, opened 2021-01-19
Autorefresh the Public and Channel screens

KyleMaas:
When the browser is scrolled to the top, autorefresh the Public and Channel screens.  They refresh faster when there are no messages showing (such as when a new profile is first being set up and it's doing the initial sync), but then it refreshes every 30 seconds for as long as the browser is left scrolled to the top.  Once you scroll down, autorefresh is disabled.

### DIFF
--- a/ui/channel.js
+++ b/ui/channel.js
@@ -32,6 +32,7 @@ module.exports = function () {
         offset: 0,
         pageSize: 50,
         displayPageEnd: 50,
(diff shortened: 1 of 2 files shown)


arj03:
I did actually design the public thread so it wouldn't update automatically because I find the little icon showing new messages to be less intrusive, but I guess if this only updates when you are at the top then I'm okay with this :)

PR #84, opened 2021-01-20
New theme - Floral Garden Bird

KyleMaas:
Another new theme.


arj03:
Thanks, I wonder how we can make it easier to add these new themes. This one includes a font which seems to be always loaded so that's not really going to scale.

KyleMaas:
According to the CSS spec on font loading:

https://www.w3.org/TR/css-fonts-3/#font-face-loading

> User agents that download all fonts defined in @font-face rules without considering whether those fonts are in fact used within a page are considered non-conformant.

When I tried it in both Chromium and Firefox, neither one attempts to download the font if you do not have that theme active.

arj03:
Alright, I did not now that. I learning something today. Thanks :)

PR #86, opened 2021-01-20
Filter by favorite channels

KyleMaas:
Adds to #80 (so I could reuse the channel list functionality and only put that list together once) to allow for filtering the Public tab by favorite channels.  The channel list is saved separately from whether the filter is active or not so that you can add channels to the list and have them preserved across page loads regardless of whether you have the filter active.

This also fixes a bug where filter preferences were not being saved properly.

Also fixes an exception that could be thrown from ssb-msg if a message was not formatted properly and did not have content.

### DIFF
--- a/css/public.css
+++ b/css/public.css
@@ -5,3 +5,8 @@
 #onboarding-dialog {
   position: absolute;
 }
+
+#public .channel-selector {
+  float: right;
+  width: 15rem;
+}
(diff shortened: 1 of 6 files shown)


arj03:
Merging to see how this looks.

arj03:
Cool, looks nice :)

PR #88, opened 2021-01-20
Add support for internationalization

KyleMaas:
* Adds support for internationalization based on the browser's locale, if available.
* Moves strings to an internationalization dictionary in messages.json.
* Adds a UK English translation.
* Adds a setting to override the browser's locale under Settings.

### DIFF
--- a/browser.html
+++ b/browser.html
@@ -24,24 +24,24 @@
     <div id="app">
       <h1>{{ appTitle }}</h1>
       <p id="navigation">
-        <router-link to="/public">Public</router-link><new-public-messages></new-public-messages> |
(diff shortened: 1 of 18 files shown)


arj03:
Ok I guess. Is messages.json missing from this PR? Looks like it needs to be included?

KyleMaas:
Huh.  You're right.  Should have it now.

arj03:
@KyleMaas is there a way to make [vue-i18n] less noisy by default? I get a ton of warnings in my console with this enabled :)

KyleMaas:
I can try.  What is your browser's default locale set to (navigator.language)?

arj03:
Its set to "en-GB"

PR #89, opened 2021-01-21
Add page titles for history

### LINKED ISSUE #65: Proper page titles for each action
When navigating between user profiles, threads, messages, etc. each action creates a new history item in the browser.  Unfortunately, they're all the same, so if you pull up the Back or Forward menus to navigate back/forward to a particular spot, you can't really distinguish between them.  By changing the title at each step, the experience of navigating using traditional browser navigation methods would be much improved.

KyleMaas:
Add proper page titles so they show up meaningfully in browser history.  Based off of the localization feature (#88).#88 

Fixes #65.

### DIFF
--- a/browser.html
+++ b/browser.html
@@ -24,24 +24,24 @@
     <div id="app">
       <h1>{{ appTitle }}</h1>
       <p id="navigation">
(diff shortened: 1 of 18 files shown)


arj03:
Looks good thanks! I'll merge and do a little local testing.

Issue #67, opened 2021-01-18
Thread/post titles


KyleMaas:
As I understand it, there is at least one SSB client which allows you to add titles to posts which are stored in an attribute somewhere in the post data (don't know if it's a root-level attribute or under content or something).  I don't believe this is in the standard for the protocol.

However, one way or another, it would be nice if at least threads either showed a title from the root post (if it's available) or tried their best to generate one based on the root post's content.  This could ideally be used for both the title of the Thread view as well as in the "replied to" links within individual posts everywhere else.

KyleMaas:
So it looks like this is where that title comes from:

https://github.com/ssbc/patchwork/blob/master/lib/depject/message/async/name.js

KyleMaas:
I'm not sure I'd consider #93 to be a full fix of this just yet.  Thinking if we're going to show them, we should probably also support posting them, with a warning to the user that not all clients support it.

KyleMaas:
I think between #93, #95, and #96 we could probably call this one fixed.

arj03:
Yes :)

PR #93, opened 2021-01-21
Show thread title on thread page, if applicable

KyleMaas:
Displays a thread title, if there is one, on the Thread page.  See #67.


arj03:
Thanks, looks good :)

PR #94, opened 2021-01-21
Silence i18n fallback warnings

KyleMaas:
If i18n has to use a fallback locale, this disables the warnings.  See https://github.com/arj03/ssb-browser-demo/pull/88#issuecomment-764677985


arj03:
Sweet!

PR #95, opened 2021-01-21
Display thread titles in ssb-msg

KyleMaas:
Displays thread titles in ssb-msg, which means that "in reply to" now shows meaningful information about what they replied to.  Tries to load from cache first, and if it's not there, asynchronously requests the thread root be downloaded so it can display the thread title.


arj03:
nice

PR #96, opened 2021-01-21
Allow adding title when starting new thread

KyleMaas:
When posting a new thread on Public, this allows you to add a (probably nonstandard, but some clients support it) title to the thread.  There is a warning to users that not everyone will be able to see it.


arj03:
Looks good.

PR #97, opened 2021-01-21
Add friendly names for existing connections

KyleMaas:
Adds friendly names for existing connections in the Connections screen.

### DIFF
--- a/ui/connections.js
+++ b/ui/connections.js
@@ -146,6 +146,32 @@ module.exports = function () {
     },
 
     watch: {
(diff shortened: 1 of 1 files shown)


KyleMaas:
And that should give us friendly names for existing connections, too.  At least as much as is possible before we are able to load anything.

arj03:
The via room-endpoint is a really nice touch. Love it!

KyleMaas:
Thanks!

PR #99, opened 2021-01-22
Add SSB.getProfileNameAsync() so we can change implementation

KyleMaas:
Adds a new SSB.getProfileNameAsync() function.  It works similarly to SSB.getProfileAsync(), except it only returns a name.  That way we can use other, possibly better and faster, ways of implementing that functionality.

I figure it's better to change things to the SSB plugins in one place rather than change each of them individually.

### DIFF
--- a/net.js
+++ b/net.js
@@ -148,6 +148,17 @@ SSB.getOOO = function(msgId, cb) {
   })
 }
 
+SSB.getProfileNameAsync = function(profileId, cb) {
+  // Get only the name from a profile.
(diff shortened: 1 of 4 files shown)


arj03:
Looks great. A small note all of these functions running on a single profile I would suggest just to use ssb-about directly. And only suggest for the private compose.

KyleMaas:
I was planning on using ssb-social-index directly for that.  Would ssb-about be more appropriate?  We don't currently use that as a dependency, AFAICT.

PR #108, opened 2021-01-23
Fix problems with popups in themes

### LINKED ISSUE #102: Ethereal theme popups are nearly invisible
Text in popups, such as message posting previews, key exports, etc. is nearly invisible in the Ethereal theme because of the white backgrounds with white text.

KyleMaas:
Fix problems with light text on light backgrounds in themes which use light text as their default.  Fixes #102.


arj03:
Easy merge. Thanks

PR #109, opened 2021-01-23
Move profile page information around

### LINKED ISSUE #107: On Profile pages, display profile's name near the top
It would be really nice if there was a spot near the top of the Profile page that displayed the person's name, and not just their description.  Like when you click on your own profile and you get the text box near your profile picture.

For that matter, the description should probably be moved up so it's more consistent with the layout of the profile page for yourself, with the name and description both placed above the action buttons.

KyleMaas:
* Make the "self" profile layout more similar to the "others" layout.
* Use space a little more effectively to put more information closer to the top of the page and above the fold.

Fixes #107.

### DIFF
--- a/css/profile.css
+++ b/css/profile.css
@@ -1,14 +1,44 @@
+#profile > span {
+  display: flex;
+}
+
+#profile > span > .avatar {
+  flex: none;
+  width: 13rem;
+  margin-right: 1em;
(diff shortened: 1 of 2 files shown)


arj03:
Merging to see how it looks

arj03:
Looks really nice. Thanks :)

PR #110, opened 2021-01-23
Stop the autorefresh timer when not on Public

### LINKED ISSUE #98: Disable autorefresh when navigating to other tabs
If you go to the Public tab and it installs the autorefresh timer, it does not clear it when you leave the Public tab, so it keeps trying to refresh even when on other tabs.

KyleMaas:
Disable the autorefresh timer when the user navigates away from the Public tab.  Fixes #98.


arj03:
Thanks

Issue #103, opened 2021-01-23
More translations


KyleMaas:
We have an internationalization framework.  Now, in the interest of making this more universally-accessible, we need some more translations.

arj03:
Yes! A pirate translation would be nice :)

KyleMaas:
> Yes! A pirate translation would be nice :)

This just became an absolute necessity.  :)

KyleMaas:
@arj03 If you know any other languages, it'd be great if we could add more.  I know a very small amount of Spanish (about four years of classes in my teens) and would have to lean heavily on machine translation there.  I more or less gave up on trying to learn French, but that'd be a good one to add, too.

arj03:
The important thing is that the feature is there now. More can be added as people of different languages start using the program. Might be good to add a section to the README about how to do that.

PR #111, opened 2021-01-24
Add translations

KyleMaas:
* Fix a couple typos in the base English translation.
* Add Japanese translation based mostly on machine translation with some minor tweaks from my not-even-N5 level of Japanese proficiency.
* Add Pirate translation, per https://github.com/arj03/ssb-browser-demo/issues/103#issuecomment-766156259

### DIFF
--- a/messages.json
+++ b/messages.json
@@ -57,7 +57,7 @@
       "possibleConnections": "Possible connections",
       "connectToX": "Connect to {peer}",
       "existingConnections": "Connections",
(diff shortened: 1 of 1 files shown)


KyleMaas:
I will note that I'm sure the Japanese will need to be corrected.  It's a start.  I have a passing familiarity with really basic Japanese, but it's nowhere near that of a native speaker, and the technical level of a lot of our strings is way above my skill level.  So I'm hoping the machine translation did a reasonable job, although I did correct some of what looked like really bad translations.

arj03:
Great thanks

arj03:
Lol this pirate translation is golden. It seems like there are quite a few things that still needs translating?

KyleMaas:
By all means, please add more!  I was running out of ideas.

PR #112, opened 2021-01-24
Private message validation

### LINKED ISSUE #87: Private message recipients are not validated properly
You can post private messages to "This is invalid" and it does not complain.

KyleMaas:
Fixes issue where you could freeform type in a name into the Private Message recipient field and proceed to post a message without any complaints.

This pull request is based on my branch for Japanese language support (pull request #111) so I could add the error message in both English and Japanese (machine-translated).

Fixes #87.


arj03:
Looks good, thanks

Issue #32, opened 2021-01-14
Timeline improvements


arj03:
So loose ideas:
 - Grouping of people
 - [Popular posts](https://github.com/fraction/oasis/blob/71926c7bd53a2c031a9c3d1af47d4bb0aa94629e/src/index.js#L441)

KyleMaas:
How would grouping of people work?

arj03:
Maybe something like one of these:
 - Write lists in local storage
 - Write lists as private messages to yourself. Not here there is a limit here in message size, so maybe as a blob if the lists gets large enough

arj03:
A much easier thing than the group might be to have a direct follow filter like the threads filter.

KyleMaas:
That much, I could accomplish.  See #50.

KyleMaas:
From here:

https://github.com/fraction/oasis/blob/d6737020110655b1b2a0c0903ea2746d02dc1966/src/models.js#L1294

It looks like they're determining popularity by quantity (and value, positive or negative) of votes within a recent time period?

arj03:
@KyleMaas yes. I did an experiment once where the user would be able to fiddle with the variables: how much weight is votes / posts from friends compared to others. I think for now it would be best with a simple algorithm like the one in oasis :)

PR #113, opened 2021-01-24
Add a filter to Public to allow for hiding of channels

KyleMaas:
If you want to ignore a channel, this adds a filter for that.  Does not hide the unwanted channels from the Channels tab, since they should still be accessible to the user if chosen.

This is based on the pull request #111, since I wanted to make sure there was a machine-generated translation of the filter description.

Related to #32.

### DIFF
--- a/css/public.css
+++ b/css/public.css
@@ -10,3 +10,7 @@
   float: right;
   width: 15rem;
 }
+
+#public .filter-line {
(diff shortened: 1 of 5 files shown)


arj03:
This is a nice feature. Code looks good from looking. Will merge and see how it behaves.

Issue #114, opened 2021-01-24
Something is strange with the auto refresh


arj03:
I'm getting a bunch of Refreshing in the console, about 1 per second and the UI grounds to a halt.

KyleMaas:
Is this on an existing profile or starting from scratch?

arj03:
Existing. 

Seems to happen if you are navigate to lets say profile and go back to public. I added some debug and I could see:

```
adding timer
Refreshing
refreshing from auto timer!
adding timer
Refreshing
refreshing from auto timer!
adding timer
Refreshing
```

KyleMaas:
I'm thinking it must be setting a timer when there already is one.  Since it only tracks one at a time, it would never clear the other one(s).  I'll see what I can do.

KyleMaas:
@arj03 Attempted fixes in #115.  I can't seem to duplicate this problem, so you'll probably need to try it and see if it works.

KyleMaas:
I think #119 would fix this, but it needs to be tested.

KyleMaas:
> Seems to be working much better now

From @arj03 here: https://github.com/arj03/ssb-browser-demo/pull/119#issuecomment-769845708

I'm assuming this can be closed now.

PR #115, opened 2021-01-24
Autorefresh changes

KyleMaas:
* Attempt to fix the problem reported in #114.
* Adds setting to turn on/off autorefresh.
* Changes timing from checking every 3 seconds when there are no messages and 30 seconds when there are to 5s and 60s.
* Fixes some tabs in Settings that should have been space indented.


arj03:
This looks like it could fix the problem.

arj03:
It's not working. As I said when looking at the initial PR there already is a way to see if there are new messages. I didn't think this would be a big deal, but I think it's one of those things that can be quite tricky to get right and might have a lot of edge cases. So I'm inclined so simply removing the feature. I hope you won't take that the wrong way. You are doing awesome work here, I just fear this is going to be a maintenance nightmare.

KyleMaas:
Is it possible you could just set the default preference to disabled instead (defaultprefs.json as of pull request #115)? That way, I can continue to work on this.

PR #116, opened 2021-01-24
Don't try to pull thread titles from ooo

arj03:
@KyleMaas I noticed that getting these messages ooo style in order to have the title show up properly was quite draining in the public view. So I'm inclined to just only to that for local messages.

### DIFF
--- a/ui/ssb-msg.js
+++ b/ui/ssb-msg.js
@@ -227,21 +227,10 @@ Vue.component('ssb-msg', {
 
     // If it's a reply to a thread, try to pull the thread title.
     if (this.msg.key != this.thread) {
-      // Try local first.
       SSB.db.get(this.thread, (err, rootMsg) => {
         if (rootMsg) {
           var newTitle = helpers.getMessageTitle(self.thread, rootMsg)
           self.parentThreadTitle = (newTitle != self.thread ? newTitle : self.$root.$t('ssb-msg.threadTitlePlaceholder'))
-        } else {
-          // Not local. Wait until the connection comes up and try to fetch it.
-          SSB.connectedWithData(() => {
-            SSB.getOOO(self.thread, (err, rootMsg) => {
-              if (rootMsg) {
-                var newTitle = helpers.getMessageTitle(self.thread, rootMsg)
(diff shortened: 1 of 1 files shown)


KyleMaas:
This would only really make a difference for when a thread root is posted by someone you're not following, correct?

KyleMaas:
Otherwise it would always be local, because partial replication is only used for hops >1?

arj03:
We get full messages for the ones you are following and partial messages for what they are following. So it would only be messages outside this or older messages. I just noticed that it fired of maybe 20 ooo request and that seemed excessive.

KyleMaas:
Not being familiar with how everything underneath this works, is there a way to batch those into one deferred request?

arj03:
Not right now. We might be adding that later.

KyleMaas:
Main reason I added this is that "Replied to a thread" doesn't provide any context for replied messages, and it's hard to know which messages in the timeline can be mentally grouped into the same topic as you're reading them.  Or to know if a message is part of a thread you've seen before, and therefor something to pay attention to.  Although if you've seen it before, it would be local, so maybe that's not as much of a concern.  I guess we could do this for now and look at reintroducing it when batched requests are supported.

arj03:
Exactly :)

KyleMaas:
I will say at this point, having tried this, that nearly all of the messages in Public are now displaying as "a thread" with no context.  Maybe this could be reintroduced as a preference, with the default off?

KyleMaas:
Out of the first 20 replies (excluding thread roots) in my Public tab, 5 of them have "replied to" titles, and 15 are just listed as "a thread".

arj03:
That sounds a bit wierd. Is that on you new Kyle profile?

You should be getting the latest 25 messages of feeds followed by feeds you follow, so it sounds strange that you are missing to many. What I want to say it that this sounds more like a bug in that part.

The code that was removed had at least two problems: 1) it would just queue up messages so if you were offline for an extended amount of time you potentially had a lot of those 2) it doesn't store the ooo messages, meaning you download them every time you go to the public page. There really should be 2 ooo methods (peek and get + store).

I would go looking for why you are missing so many messages first of all.

PR #117, opened 2021-01-24
Remove autorefresh feature

arj03:
This removed the feature as argued [here](https://github.com/arj03/ssb-browser-demo/pull/115#issuecomment-766419729)


KyleMaas:
Is it possible you could just set the default preference to disabled instead (defaultprefs.json as of pull request #115)?  That way, I can continue to work on this.

arj03:
I could, it needs to be applied a few more places then. Just trying to feel how much you really like this feature? ;)

KyleMaas:
I actually use it quite a bit.  It's nice for being able to passively watch Public while working on something else.

arj03:
Alright, will do that instead then :)

KyleMaas:
Thanks!

arj03:
I pushed up the disable now

KyleMaas:
Cool.  That should work.

PR #118, opened 2021-01-24
Add new theme: Pirate Ship

KyleMaas:
We have a translation.  We might as well have a theme.


arj03:
Hah, the font is maybe a bit over the top. Anyway love painting my ship with this, mate!

KyleMaas:
Well, "subtle" and "understated" were not exactly terms I had in mind when making this.  :)

PR #119, opened 2021-01-25
Event-driven autorefresh on Public tab

KyleMaas:
Instead of figuring out autorefresh on the Public tab, this version does it the new-public-messages component.  The idea is that when a new message comes in, it checks to see if autorefresh is enabled and, if so, refreshes.  But then it blocks autorefreshing for 30 seconds.  At the end of that timer, it rechecks in case a message came in while it was blocked, and then it unblocks autorefreshing so the next message can refresh immediately.

The reason for the timer is because when you're doing a major sync operation it could otherwise trigger a whole bunch of refresh requests in a very short amount of time (it did this the first time I tried it).  So I think we still need a timer, but it needs to block instead of being the primary motivation for action.

So, does that look like an acceptable way to do it?

### DIFF
--- a/ui/new-public-messages.js
+++ b/ui/new-public-messages.js
@@ -48,7 +48,7 @@ module.exports = function (state) {
               self.newPublicMessages = true
 
               // If we're scrolled to the top of the page and autorefresh is on, refresh.
-              if (self.publicRefreshTimer == 0) {
+              if (self.publicRefreshTimer == 0 && localPrefs.getAutorefresh()) {
                 // Only allow refreshing every 30 seconds, but at the end of that, check once again if we have queued messages.
                 self.publicRefreshTimer = setTimeout(function() {
                   console.log("Checking for new data via timer...")
--- a/ui/new-public-messages.js
+++ b/ui/new-public-messages.js
@@ -1,6 +1,7 @@
 module.exports = function (state) {
   const { and, type, live, toPullStream } = SSB.dbOperators
   const pull = require('pull-stream')  
+  const localPrefs = require('../localprefs')
 
   var loaded = false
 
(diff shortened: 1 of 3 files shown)


KyleMaas:
I think this should fix #114

KyleMaas:
Does this look like a reasonable way to do it?

ui/new-public-messages.js
```diff
@@ -36,8 +44,26 @@ module.exports = function (state) {
           live(),
           toPullStream(),
           pull.drain((msg) => {
-            if (!msg.value.meta)
+            if (!msg.value.meta) {
               self.newPublicMessages = true
+
+              // If we're scrolled to the top of the page and autorefresh is on, refresh.
+              if (self.publicRefreshTimer == 0) {
```
arj03, line 51: this needs to be only enabled if the settings is true. We don't want to do anything unless that is set.

ui/public.js
```diff
@@ -276,21 +261,25 @@ module.exports = function (componentsState) {
...
+          return
+
+        this.isRefreshing = true
+
         console.log("Refreshing")
         this.messages = []
         this.offset = 0
-        this.renderPublic()
+        this.renderPublic((err, success) => {
+          // Clear the refreshing flag.
```
arj03, line 274: Remove this comments, it is pretty clear what is going on

ui/new-public-messages.js
KyleMaas: That's checked inside of refreshIfConfigured.  It's checked there because new-public-messages is instantiated once and doesn't get reloaded as you change between tabs, so if you turn off the setting and save, the timer might still autorefresh once if it was checked outside of refreshIfConfigured.

KyleMaas:
Changed

ui/new-public-messages.js
arj03: Right, but I don't want to to do anything if not configured. Will be much easier to figure out if its working or not.

KyleMaas: See latest commit.

arj03:
Thanks :)

KyleMaas:
Have you had a chance to try this out yet and see how it works for you?  Because if this method is suitable, I'd like to try to apply it to Private, Thread, and Channel as well.  If it isn't, I'd like to file an issue so I know what needs work.

arj03:
Seems to be working much better now

KyleMaas:
Cool.  Then I will see if I can get this implemented on the other pages for consistency.

PR #121, opened 2021-01-25
Use ssb-social-index for names, if available

KyleMaas:
This seems to actually be working now.


net.js
```diff
@@ -151,48 +151,55 @@ SSB.getOOO = function(msgId, cb) {
...
+    if (err) {
+      console.log("Got error from ssb-social-value: " + err)
       cb(err)
-    else
-      cb(null, profile.name)
+    } else if (value) {
+      console.log("Got name from ssb-social-value: " + value)
+      cb(null, value)
+    } else {
+      // Fall back to pulling it from the server.
```
arj03, line 163: This shouldn't be needed. getProfileAsync will also use SSB.net.about.socialValue.

net.js
```diff
@@ -151,48 +151,55 @@ SSB.getOOO = function(msgId, cb) {
...
-            })
-          )
+    // Going to have to fetch it when a connection comes up.
+    SSB.connectedWithData(() => {
+      let rpc = SSB.getPeer()
+
+      if(!rpc) cb("No peer to connect to")
+
+      pull(
+        rpc.partialReplication.getMessagesOfType({id: profileId, type: 'contact'}),
```
arj03, line 190: This is a bit strange. We are fetching contact messages for the profile in question, but socialValues uses about messages.

arj03:
I'll rework this a bit locally while testing. Really great work!

KyleMaas:
Please, by all means.  I can't say I'm particularly happy with what I was able to do with this.

arj03:
It's a lot easier to tweak something that is almost working than having to go through all of this myself. I'm still testing and revising.

KyleMaas:
Sounds good.  Would be great if you could file a few more issues sometime for things you'd like to see worked on.

arj03:
Will do, there was a super awesome feature in patchbay that I really miss. Will open an issue for that a little later.

arj03:
Continued [here](https://github.com/arj03/ssb-browser-demo/pull/123). Thanks for your work on this @KyleMaas

PR #122, opened 2021-01-25
Disable autorefresh by default

KyleMaas:
Disables autorefresh by default unless specifically enabled under Settings.


arj03:
Missed that. Thanks :)

PR #123, opened 2021-01-26
Ssb about suggest

arj03:
Continuation of https://github.com/arj03/ssb-browser-demo/pull/121

### DIFF
--- a/net.js
+++ b/net.js
@@ -173,39 +173,33 @@ SSB.getProfileNameAsync = function(profileId, cb) {
 }
 
 SSB.getProfileAsync = function(profileId, cb) {
-  const profiles = SSB.db.getIndex('profiles').getProfiles()
+  const keysWeWant = ['name', 'description', 'image']
+  SSB.net.about.socialValues({ keys: keysWeWant, dest: profileId }, (err, values) => {
 
-  if(profiles[profileId]) {
+  if(values) {
     // We've got it already.
-    cb(null, profiles[profileId])
+    cb(null, values)
   } else {
-    // Don't have it yet.  Let's try loading the graph.
-    SSB.db.getIndex('contacts').getGraphForFeedHops1(profileId, (err, data) => {
-      const profiles = SSB.db.getIndex('profiles').getProfiles()
+    // Going to have to fetch it when a connection comes up.
+    SSB.connectedWithData(() => {
+      let rpc = SSB.getPeer()
 
-      if(!err && profiles[profileId]) {
-        // Got it from graph.
-        cb(null, profiles[profileId])
-      } else {
-        // Going to have to fetch it when a connection comes up.
-        SSB.connectedWithData(() => {
-          let rpc = SSB.getPeer()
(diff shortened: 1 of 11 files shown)


KyleMaas (commented):
I originally put that in to support the "Send Message" buttons from the Profile pages.  Without that, the page would load but wouldn't have a person in the options list for the v-select to work with.

arj03:
This is a bit closer to being merge ready. I'm out of time for today.

Some things that would be nice to have fixed:
 - figure out a way to get images in private (I tried doing it async, but I think the list gets populated, so maybe some kind of smart cache)
 - speaking of cache, it actually might make a lot of sense to have profile cache
 - this might fix the issue with vote names that are also missing

Suggest seems to have a rather large [timeout](https://github.com/ssbc/ssb-suggest/blob/1d307ece57e84ee3643b684bb4b6463861fd5784/index.js#L31). Might be interesting to play around with if we really need that large of a timeout when using db2.

arj03:
Implemented a naive cache and used that in places where it makes sense. This should be ready to go once we have ssb-suggest merged. 

@KyleMaas can you give it a span?

KyleMaas:
Sure.  Give me a few minutes - trying to get the progress on the Markdown editor pushed.

KyleMaas:
Odd exception I got when trying to open the recipients box in Private:

```
bundle-ui.js:105715 Uncaught (in promise) Error: Input must be string
    at sanitize (bundle-ui.js:105715)
    at module.exports (bundle-ui.js:105728)
    at sanitizedPath (bundle-ui.js:113315)
    at localGetHelper (bundle-ui.js:113571)
    at Object.localProfileGet (bundle-ui.js:113662)
    at Object.hooked [as localProfileGet] (bundle-ui.js:71472)
    at bundle-ui.js:132724
    at bundle-ui.js:121693
    at bundle-ui.js:100159
    at bundle-ui.js:100504
```

KyleMaas:
Thoughts:

1. That works considerably better than my version.  I wasn't happy with the way it worked before.  This is pretty good.
2. Names take a few seconds to load when I first refresh the page.  But they are able to load without having to connect to a peer, so that isn't too terrible.
3. Kind of wish the Recipients for private messages had placeholder images for folks who don't have a profile image.  Or even just a blank space.  Just something so that the text in their names lines up with the folks with images.

Overall, this is a major improvement over my version, and from a user experience standpoint seems to be pretty close to the same experience as we had before.  So if the performance is better, then "externally relatively close to the same" is good.  :)

KyleMaas:
Oh, and reaction names do seem to be working now.

KyleMaas:
Glad to see that off of my fork.  That was beginning to cause some build problems for me.

arj03:
I also got the Input must be string error now

arj03:
@KyleMaas regarding "Kind of wish the Recipients for private messages had placeholder images for folks who don't have a profile image". Yes that's a really good idea, is that something you can do?

PR #125, opened 2021-01-26
Add DHT Support - Runnable without manual patching

KyleMaas:
Adds DHT support.  This one is runnable without manual patching of dependencies.

Please note that this is not intended to actually be merged - too many things are pulled in from my own forks for the modifications I had to make.  It would work, but I'd much rather see upstream fix their issues instead.


KyleMaas:

PR #128, opened 2021-01-27
Markdown editor

KyleMaas:
This adds the Markdown editor from toast-ui for posting messages and profile descriptions.

(I was unable to test profile descriptions properly due to issues between the current master branch of ssb-browser-demo relying on the older version of ssb-browser-core with the "profiles" index.  But the code is very similar to the message code which I was able to test.)


KyleMaas:
Apparently I still had it using my local checkout of ssb-browser-core.  Tested it using released version of ssb-browser-core instead.  Profile editors seem to mostly work, as long as it's able to load the profile, which on the old version apparently doesn't always work when you haven't synced with a pub yet.  So...yeah.  Pretty sure this works.

I will note that image embedding embeds using a data: URL instead of attaching as a blob.  That can be worked on, but I wanted to see if I was going in a good direction with this first.

arj03:
Markdown editor is a good idea but we need the blobs part working before merging this. So please continue on this :-)

KyleMaas:
Understandable.  But I wanted to make sure I was heading in a sane direction before putting a whole lot of time into this.  :)

arj03:
I hear you, great you are reaching out first :heart:

KyleMaas:
Looks like this'll be easier than I thought:

https://github.com/nhn/toast-ui.react-editor/issues/7

KyleMaas:
Better guide:

https://github.com/nhn/tui.editor/issues/1280

I have this partially working.  Just need to hook the preview so the image is shown correctly in WYSIWYG mode.

KyleMaas:
Also, this is pertinent - we can't use the alt text from the insert dialog for now:

https://github.com/nhn/tui.editor/issues/1271

KyleMaas:
That should be pretty well working.  Go ahead and give it a try.

arj03:
From the code this looks really nice. I think I'll merge in the ssb-about / ssb-suggest first. Should land ssb-suggest today so that we can create a new release of ssb-browser-core. One question, what is the file size of this after compared to before for the dist bundle file?

KyleMaas:
Release build goes from about 4240kb on master branch to 5026kb.

arj03:
Can you rebase this against latest master now that the ssb-about stuff is merged? Thanks :)

arj03:
And just saw this in console :)

[BABEL] Note: The code generator has deoptimised the styling of /home/arj/dev/kyle-test/node_modules/@toast-ui/editor/dist/toastui-editor.js as it exceeds the max of 500KB.

KyleMaas:
Yeah, I saw that Babel note, but after looking into it before it appears it's because the editor is already prepackaged/bundled, so it probably wouldn't buy us much anyway to change that.  I've tried the recommended "fix" for that of adding a .babelrc file to tell it to ignore that, but it didn't seem to work.

arj03:
Was testing this. Awesome work!

KyleMaas:
Thanks!  Glad you like it.

PR #133, opened 2021-01-30
Profile image placeholders on private

### LINKED ISSUE #131: Profile image placeholders in Private recipients
> @KyleMaas regarding "Kind of wish the Recipients for private messages had placeholder images for folks who don't have a profile image". Yes that's a really good idea, is that something you can do?

From @arj03 here: https://github.com/arj03/ssb-browser-demo/pull/123#issuecomment-769721625

KyleMaas:
* Centralize "blank profile" image into helpers so it can be reused in multiple places without inline builds adding multiple copies of it.
* Add "blank profile" image to Private tab's Recipients list for people who don't have a profile image.
* Sort Recipients suggestions by name.
* Don't wipe out the people list in suggestions if no matches are found, also making it consistent with how recipientsOpen() works.
* Only update the component's people list once per method instead of modifying it (and possibly triggering hooks/processing) every time an element is pushed onto the array.
* Clean up a few stray tabs.

Fixes #131.

### DIFF
--- a/ui/helpers.js
+++ b/ui/helpers.js
@@ -40,3 +40,9 @@ exports.getMessageTitle = function(msgId, msg) {
   // Fallback - use the message ID.
   return msgId
 }
+
+exports.getMissingProfileImage = function() {
+  // This is centralized here so that when building to a single inlined HTML file,
+  // we're only swapping in the base64-encoded version once.
+  return "assets/noavatar.svg"
+}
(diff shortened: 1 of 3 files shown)


arj03:
This looks good. I'll just note that ssb-suggest does sort the results so that the most relevant are supposed to be first.

KyleMaas:
Ah, okay.  I did that because when you go to Private, start a new message, and click into the Recipients box to pop it open, they looked like they were in semi-random order.

arj03:
Yeah, it is a bit confusing

arj03:
Think it's good as you did it

KyleMaas:
Sounds good.

PR #135, opened 2021-01-30
Make sure there's a message before trying to post it

### LINKED ISSUE #127: Missing empty message validation for "Channel" and "Public"
"Private" validates new messages to make sure they actually contain something (not just an empty string for a message).  We should be validating that for "Channel" replies and "Public" posts as well.

KyleMaas:
Validates that the message field is not empty when trying to post on both Public and Channel.

Fixes #127.


arj03:
nice

PR #136, opened 2021-01-30
Enforce message size limits

### LINKED ISSUE #130: Enforce size limits on messages
Need to make sure that when we go to post a message that the size doesn't exceed the maximum per-message size for Scuttlebutt messages - 8KB including headers, according to:

https://scuttlebot.io/docs/message-types/post.html

KyleMaas:
Warn the user if their message goes over the 8192 byte limit, rather than just fail with a message on the console.

Fixes #130.


arj03:
Looks good, there is a small conflict now :)

arj03:
Lets see

arj03:
Botched that, will fix after merge

PR #143, opened 2021-01-30
Include replies in Notifications

KyleMaas:
Attempted to do this using pull-stream.  I think it seems to work okay, although I'm also not an active enough poster to really give this a good test.

### DIFF
--- a/ui/notifications.js
+++ b/ui/notifications.js
@@ -49,6 +49,8 @@ module.exports = function () {
       render: function () {
         var self = this
 
+        const contacts = SSB.net.db.getIndex('contacts')
+
         pull(
           cat([
             // Messages directly mentioning the user.
@@ -73,8 +75,8 @@ module.exports = function () {
           self.getSameRoot,
           pull.unique('key'),
           pull.filter((msg) => {
-            // Exclude messages from user.
-            return (msg.value.author != SSB.net.id)
+            // Exclude messages from user and blocked authors.
+            return (msg.value.author != SSB.net.id) && !(contacts.isBlocking(SSB.net.id, msg.value.author))
           }),
           pull.collect((err, msgs) => {
             // Only show the most recent 50.
--- a/ui/notifications.js
+++ b/ui/notifications.js
@@ -2,7 +2,7 @@ module.exports = function () {
   const pull = require('pull-stream')
   const cat = require('pull-cat')
 
-  const { and, mentions, author, type, toCallback, toPullStream, hasRoot } = SSB.dbOperators
+  const { and, mentions, author, type, toCallback, toPullStream, hasRoot, descending } = SSB.dbOperators
   
   return {
     template: `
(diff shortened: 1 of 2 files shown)


KyleMaas:
One thing I can think of that we might want to change about this is to "clear" notifications after you react, just like how it excludes posts in threads from before when the user posted to them.  But I wanted to check this over first and see if I'm going in a valid direction, and besides, improvements could be made in other pull requests.

KyleMaas:
Hm...do we need to also exclude blocked users at the UI layer here, or would that be already filtered out underneath somewhere?

ui/notifications.js
```diff
@@ -17,11 +20,56 @@ module.exports = function () {
...
-        SSB.db.query(
-          and(mentions(SSB.net.id)),
-          toCallback((err, results) => {
-            this.messages = results
+        var self = this
+
+        pull(
+          cat([
+            // Messages directly mentioning the user.
+            SSB.db.query(
```
arj03, line 54: Since we are getting roots for both of these, they should be capped at 25 I think. Like [this](https://github.com/ssbc/ssb-suggest/blob/027578f6afc4e05cafffb99bbec22b2b9f639c00/lib/watch.js#L115).

arj03:
I tried testing this but didn't get any results. I just replied to one of my browsers old threads and that message should show up as a notification. The good thing about that code is that its easy to change SSB.net.id to some other id :-)

arj03:
Regarding blocked, then yes and no. If you block and delete then they are gone. But if you just blocked, then they need to be filtered out at the UI level. This is why I'm sort of in favor of the block & delete.

KyleMaas:
Feel free to change this if you can make it work.  I'm still not very familiar with either db2 or pull-stream, and this uses both.  It worked for me, but like I said, I don't post enough for much to actually show up, so it's hard for me to test this myself.

arj03:
Not sure I understood the clear notifications question. It is true that once you reply to a thread again that you "clear" old notifications. This might be a bit counter intuitive. Is that what you mean? I can see the implementation would be even more tricky because of this. I'm actually quite impressed with the pull thing you pulled together, it should work from reading the code at least :)

KyleMaas:
The question about "clearing" notifications is that I would think that reacting to a post would indicate that you've read something just as much as replying to a thread would.  Do we want to filter out notifications for messages prior to a reaction (what I was asking about) rather than just messages prior to a reply (how it works now)?

And thanks!  It was tough, trying to learn the pull-stream API at the same time, but I figured if I could implement it that way, then we could use something similar to display a "new notifications" indicator and also do event-driven autorefresh.

arj03:
I think it's fine how it works in this PR regarding clearing notifications

arj03:
Found the problem. When you open a thread, root is not set. So you need something like this instead in getSameRoot:

```
if (data) {
   const root = data.value.content.root ? data.value.content.root : data.key
```

With that I got my reply :)

KyleMaas:
How's that look?

arj03:
Thanks :)

PR #144, opened 2021-01-30
Add Markdown editor to Thread view

### LINKED ISSUE #138: Thread view needs Markdown editor
Apparently I missed adding the Markdown editor to Thread.

KyleMaas:
Changes the Thread view's reply area to a Markdown editor, just like Public, Private, and Channel.

Fixes #138.

### DIFF
--- a/messages.json
+++ b/messages.json
@@ -147,6 +147,7 @@
     "thread": {
       "title": "Thread \"{title}\"",
       "postReply": "Post reply",
+      "blankFieldError": "Please provide a message to publish.",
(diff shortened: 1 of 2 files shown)


arj03:
Great. Probably would make sense to refactor some of these at some point. But this is good for now.

KyleMaas:
Agreed.  I figured I'd have to do that to get the autosuggest stuff working for typing in blob IDs and mentions.  But those are going to take me a while to figure out, so I figured this would at least make it a little more usable in the meantime.

PR #146, opened 2021-01-31
Refactor Markdown editor into its own component

KyleMaas:
Refactor of the Markdown editor and associated blob adding functionality into its own component.


arj03:
This is really nice. Great to see vue components to good use here.

KyleMaas:
Yeah, it's how it probably should have been implemented originally.  But when I started on this, I didn't realize how many places this would be used.  Was better to get it going with minimal effort to see if it would even work at all.  But with trying to add suggestions (for @ mentions and blobs), it would have started to make even more of a mess.  So it needed the refactor first.

PR #147, opened 2021-01-31
Suggestions for profiles and channels

### LINKED ISSUE #129: Search profiles feature
Related to #124, it would be nice if when you started typing someone's name with an "@", it would search the contacts list for that person so you could not only mention them but include their ID as well.  And that way you'd know the spelling of their name was right, too.

KyleMaas:
This adds an extensible architecture for auto-suggesting things while the user is typing, such as profiles, channels, or (in the future) blobs.

One note on this - there is an issue where if it pops up the suggestion box and you then navigate your cursor away from where you're typing, it has no way to track that and the suggestion box stays open.  This is, unfortunately, not something I think we can fix, since we would need some kind of an event to be emitted by the TUI editor when the cursor moves.  It works without it, but just be aware that that behavior is likely not something we can fix right now.  I've reported it upstream as https://github.com/nhn/tui.editor/issues/1370

Fixes #129.

### DIFF
--- a/ui/markdown-editor.js
+++ b/ui/markdown-editor.js
@@ -1,5 +1,6 @@
 const helpers = require('./helpers')
 const ref = require('ssb-ref')
+const { and, not, isPublic, type, channel, toCallback } = SSB.dbOperators
 
 Vue.component('markdown-editor', {
   template: `<div class="markdown-editor">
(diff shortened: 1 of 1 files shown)


ui/markdown-editor.js
```diff
@@ -75,6 +76,180 @@ Vue.component('markdown-editor', {
       this.blobUrlCache[blobId] = blobURL
     },
 
+    suggestChannels: function(searchString, cb) {
+      const searchForChannel = searchString.substring(1, searchString.length)
+      SSB.db.query(
+        and(not(channel('')), type('post'), isPublic()),
```
arj03, line 82: this will look through almost the whole database. Needs some kind of limiter. We might think about doing this once and doing a channel cache. We are doing quite a lot of the same thing in /public

arj03:
Looks good, had that thing about channels I we need to fix first. But otherwise good to go.

arj03:
Also added you to this repo so it's easier to keep you branches here.

ui/markdown-editor.js
arj03: Come to think of it, I think most of these filter on channel should use the [channel operator](https://github.com/ssb-ngi-pointer/ssb-db2/blob/e3898ff578021fc8efee935252691be8947a6013/operators/index.js#L48) to search instead of streaming messages.

arj03: That one might need a:

```
 prefix: 32,
 prefixOffset: 1,
 useMap: true,
```

KyleMaas (commented):
Posted comment to the wrong place.  Please ignore.

ui/markdown-editor.js
KyleMaas: Well, that reply didn't go where I wanted.  And I had a typo.

That one does filter using not(channel('')), but I couldn't find a way to filter on only part of a channel. Or does that just mean that the channel must contain something as opposed to having to match the full string?

I will note, too, that even with that in place I still got results which had channel == null. So there may be other issues with channel().

So far, the places I know that we try to pull a channel list are:

1. Public (for the channel selector when you go to post a new message)
2. Channels tab
3. The suggestion box

I could remove the channel suggestion box for now, if you'd like.  It's not like it's a difficult thing to patch back in later.  I tried to leave this extensible enough for profiles and blobs, and then actually realized I could do channels as somewhat of an afterthought.

KyleMaas:
Joined the repo.  Thanks!

KyleMaas:
How would you go about putting together an index of channels?  I'd like to use that in #189 as well.

arj03:
For searching it would be really nice with this prefix like index search. So you type tel and get all the channels that starts with tel. I don't have time right now to go into the details of implementing it. Also this is probably a bit lower priority because its mainly for channel searching in the search box.

KyleMaas:
How about this? https://github.com/arj03/ssb-browser-core/pull/23

KyleMaas:
If we do it with an index like this, we can use it for both suggestions (in the Markdown editor and the search box) as well as for the Channels tab.

KyleMaas:
This one's obsolete with the new channel index stuff.  Closing.

PR #148, opened 2021-02-01
Suggestions for profiles only

KyleMaas:
Cherry picked just the two commits for profiles from #147.  We can always reintroduce commit 351b4c5 to bring channels back in later.


arj03:
Nice, we can always add in other stuff.

PR #152, opened 2021-02-01
Fix inlining for new resource types

### LINKED ISSUE #151: Inlined build needs to be updated
Turns out the inlining is not pulling in fonts or other theme resources, so the process needs to be updated.

KyleMaas:
We now have several SVG files, JPEG files, and fonts.  This fixes the inline build so that they're included, which means themes still work in an inlined build.

Fixes #151.


arj03:
Thanks

PR #154, opened 2021-02-02
Fix and improve Connected indicator

KyleMaas:
A bug crept into the Connected indicator with the internationalization stuff - it just flat out disappeared because I used the wrong syntax for the title attributes.  This fixes that and uses the SSB.connected() and new SSB.disconnected() functions to track the connection status instead of hooking the RPCs directly and effectively duplicating that functionality.  This also checks periodically to see if the feedSyncer is synced and uses a yellow indicator when it is connected but out of sync.

### DIFF
--- a/messages.json
+++ b/messages.json
@@ -52,6 +52,7 @@
       "loading": "Searching for channels..."
     },
     "connected": {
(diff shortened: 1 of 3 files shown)


arj03:
Great, this looks good

PR #157, opened 2021-02-02
Optimize theme images

KyleMaas:
Reduces the size of images used by themes.  In the case of Floral Garden Bird, the background size was reduced even more by using a PNG instead of SVG.


KyleMaas:
This reduces the size of the inlined build by over 2MB.

PR #160, opened 2021-02-02
Optimize inline

### LINKED ISSUE #158: Optimize inlined build
The inlined build clocks in at around 22.5MB.  That seems excessive.  About 12.5MB of that comes from the emoji font.  Almost 5MB of that comes from the font on the Floral Garden Bird theme.  The rest of the resources (even the Pirate theme's font) are pretty minimal.  But I'd like to try to optimize this a bit more.

KyleMaas:
This makes a few changes for inline builds:

1. It now shows how much space the inlined output consumes after each additional file is embedded, so space hogs can be tracked down.
2. It uses an older smaller version of Noto Emoji (from here: https://github.com/googlefonts/noto-emoji/tree/733a62212930376600d0b2fc32449f6e5a7158cb/fonts ) when inlining, to save space.  Regular builds are unaffected by this.
3. The font on Floral Garden Bird has been changed to Blocus instead of TT2020.

Altogether, this has reduced the inlined size from about 22.5MB to about 14.5MB.

Fixes #158, as much as I can for now.


arj03:
Thanks

PR #162, opened 2021-02-03
If ssb-meme is installed, use it to suggest memes

### LINKED ISSUE #124: Search blob feature
Patchbay had this super awesome feature where if you are in a input field, you could type % (I think it was) an popup would allow you to search for blobs and include in the text. This gave a really nice universe where people would tag their blobs and in this way a whole universe of expressions opened up that was potentially much more rich than just text. Not to say more fun :)

@KyleMass this was the feature I was referring to. Not sure how easy or hard it is to implement. But there is some code in patchbay somewhere to do this.

KyleMaas:
If ssb-meme (a version compatible with db2) is installed, use it to suggest memes in Markdown editors when the user types "&".

Fixes #124, assuming ssb-meme is installed.


arj03:
Thanks. Also putting this on the back burner until after the weekend :-)

KyleMaas:
No problem.  I figured that'd be the case.

KyleMaas:
I should probably put this here - this needs:

https://github.com/ssbc/ssb-meme/issues/6

KyleMaas:
Going to convert this to a draft until ssb-meme is updated.

KyleMaas:

PR #166, opened 2021-02-03
User groups

### LINKED ISSUE #165: User groups
A feature I've heard of others using on other social media platforms is the concept of grouping "friends" together (in whatever platform-specific terms they use) into a group that, when clicked on, shows all of the public posts from the folks in that group.  It would be nice if we had a way to do this.  Not private groups, but public groupings solely for the use of the user.  For example, you could have a group called "SSB Developers" which contains users you know who are active SSB developers, so you can click on that group and see all of the most recent posts from them.

I could implement this quite easily on the client side by using localPrefs.  But is there a good way to represent this as an SSB message so that groupings like this could be persistent across client database resets?  Perhaps both a private version only accessible to the user and a public version so others could access this group and import it as well?

KyleMaas:
Adds the ability to create, delete, and manage user groups.  These allow you to quickly look at messages from only certain users.

Fixes #165.  Thinking this might be the missing piece of #32 as well, but I'll leave that up to you.


arj03:
Havn't looked yet, but I'm really looking forward to giving this a spin. I'm getting close to fixing friends, but it still needs another round. The good thing is that I might have found out what the problem with the original profile/about messages was so I'm thinking of maybe going back to that if I can figure out why this is slower in the browser.

KyleMaas:
Cool.  If we could do that, then maybe we could drop ssb-suggest in favor of the local index so we don't have to worry about https://github.com/ssbc/ssb-suggest/issues/10

KyleMaas:
I will note, too, that as someone who has traditionally eschewed social media, I found this feature to be unexpectedly helpful.  It'd be great if we could figure out a way to persist this to SSB once we've run this for a bit and had a chance to shake the bugs out of it.

arj03:
This has a minor conflict now. Can you fix?

KyleMaas:
Sure.  Give me a few minutes to finish up the blocked user stuff.

KyleMaas:
That should do it.

KyleMaas:
Fixed merge conflicts again.

KyleMaas:
Were you thinking trying to get this into 4.2.0 as well, or waiting until later?

arj03:
A bit undecided on that. Was really busy with the ssb-friends and about stuff. I think it can go in. Will have a proper look today.

KyleMaas:
Okay.  I'll keep trying to keep it conflict-free, then.

css/main.css
```diff
@@ -287,6 +287,31 @@ body, html {
   background-color: #6c6;
 }
 
+<<<<<<< HEAD
```
arj03, line 290: Minor thing here

arj03:
Was reading the diff, seems like a self-contained chunk, so I'm looking forward to merging, but before I do I have a question. It seems like groups can also contain channels and not only feeds?

css/main.css
KyleMaas: Fixed in new revision.  Thanks

KyleMaas:
No, user groups only contain users.  But I put the group-making interface onto the Channels tab, since we don't really have the screen space for a "Groups" tab and the functionality of viewing a channel or a group seemed similar enough.

arj03:
Ah I see. Thanks. Will merge this tonight.

KyleMaas:
I'll be unavailable for the next several hours, so please feel free to make whatever changes to this you feel are necessary.

arj03:
OMG. This is amazing! I've wanted this feature for so long...

arj03:
We really need to make room for this somewhere in the top menu. This is so damn cool :rocket:  

I tried changing notifications to 🔔. And then moving it maybe to the right next to the other icon should leave enough room. Also at first I was expecting this to be on the public page as a filter at the top. But I think with a menu item that makes maybe more sense? I really like that you can see the members directly. I made a tiny css tweak, all of this is merged into the friends-about PR. Feel free to create a fork of that to make some tweaks.

KyleMaas:
Here 'ya go:

https://github.com/arj03/ssb-browser-demo/pull/177

It does not add filters for user groups to the Public tab.  We could maybe do that at a future date, but I'd like to have filters expand/collapse first because that filter box is already using valuable screen space and pushing down the messages.

PR #170, opened 2021-02-03
Use ssb-friends

arj03:
Depends on [PR](https://github.com/arj03/ssb-browser-core/pull/22)


KyleMaas:
Oh, nice.  I guess I didn't realize before looking at this, but this allows for hops != 1, doesn't it?

arj03:
It does, its just not 100% hooked up yet.

KyleMaas:
Awesome.  Looking forward to it.

KyleMaas:
Looking over this, I can see one thing we'll have to watch out for.  With the contacts index gone, we will need a different method of figuring out blocking status, since I believe that's based on that index everywhere that we check it.

arj03:
Yes, there should be a similar method on ssb-friends. I'll see if I can get them all in the PR.

arj03:
This needs a good round of testing. I'm going to experiment a bit with the old about index.

arj03:
Decided to do the about thing as well. There is something strange with level in the browser, it's like 40x slower. But it should be ok if we just stream the profiles instead of waiting for all.

KyleMaas:
Really excited about this!

Don't forget to merge in changes from the master branch when you do your testing - the blocked-profiles feature does have a new blocking check in it that will need to be ported to this.

arj03:
I already did :) We need to hammer out the bugs in this and then I think we can release 4.2.0

KyleMaas:
Cool.  Let me know when it's ready for testing.

arj03:
@KyleMaas it should be ready now. 

I found one bug that I think it unrelated to this PR when on a /profile page:

```
TypeError: Cannot read property 'mdEditor' of null
    at a.hideSuggestions (bundle-ui.js:165016)
    at He (vue.min.js:6)
    at a.n (vue.min.js:6)
    at He (vue.min.js:6)
    at a.e.$emit (vue.min.js:6)
    at a.forEach.e.<computed> (bundle-ui.js:18235)
    at bundle-ui.js:12991
    at forEachOwnProperties (bundle-ui.js:12306)
    at EventManager.emit (bundle-ui.js:12991)
    at MarkdownEditor._setToolbarState (bundle-ui.js:14658)
```

I feel like this is pretty good now. There are probably some more bugs, but at least this feels good. Also the API got a bit nicer. And +269 −323 is always a good sign :)

KyleMaas:
I noticed that one last night when I was working on something else.  It happened when it tried to hide a suggestions popup after having navigated to a different page when the Markdown editor was already torn down.  Should be easy enough to fix.

arj03:
It's still marked as draft until we merge the ssb-friends.

arj03:
@KyleMaas word of warning, if you update to jitdb 2.0.4 then most things will not work until you clear the indexes. Furthermore this is a [bug](https://github.com/ssb-ngi-pointer/jitdb/issues/114). We'll fix that tomorrow. So my recommendation if you want to look at some of the groups things is probably to stay on master and just make your changes there.

KyleMaas:
I ended up just cherry picking your CSS tweak commit into the User Groups Improvements branch.  Hopefully that way we won't have merge conflicts.

KyleMaas:
So with the jitdb bump, is this ready to test?

arj03:
It is :)

KyleMaas:
This is *way* faster.  UI is just overall much less laggy.  And so far, the only glitch I'm seeing is that if I look at a page with a blocked profile, I get a 404 error in the log:

GET http://127.0.0.1:8080/undefined 404 (Not Found)

So for some reason, it's coming back as 'undefined' instead of 'assets/noavatar.svg' or the actual avatar of the person.

KyleMaas:
Huh.  The actual element shows noavatar in the src.  So I'm not sure what's trying to pull `http://127.0.0.1:8080/undefined`.  But it seems to consistently happen on the profile page (at least) when it shows a blocked profile.

KyleMaas:
Ah, I think I found it.  It happens when you look at your own profile if you have an image in your description.  I think it's that the editor is trying to preview it before it's had a chance to load the image's URL into the cache.  So that's probably unrelated to this pull request.

KyleMaas:
This looks pretty good.  I've been clicking around like a madman trying to explore enough to get things to break, and so far it's just working (with the exception of that 404 that doesn't seem to be from this).

arj03:
Cool. Thanks for testing. I'll be doing a release soon :)

KyleMaas:
Yay!

PR #161, opened 2021-02-02
More reactions

KyleMaas:
Adds a popup menu with more reaction emojis.

Fixes #159.

PR #172, opened 2021-02-03
Fix positioning bug in 'More Emoji' popup

KyleMaas:
This *should* fix https://github.com/arj03/ssb-browser-demo/pull/161#issuecomment-772825937

Tried it in both Chromium and Firefox.  It was previously broken in Firefox, and now works there.  Couldn't duplicate the problem in Chromium.  But in both browsers on my machine the emojis organize into two rows.

### DIFF
--- a/css/main.css
+++ b/css/main.css
@@ -288,6 +288,11 @@ body, html {
 }
 
 /* Emoji menu */
+.message
+{
+  position: relative;
+}
+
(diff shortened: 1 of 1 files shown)


arj03:
Will try

arj03:
That did it. This is really awesome, new I can express my pizza reactions :heart:

PR #174, opened 2021-02-04
Fix mdEditor error

KyleMaas:
Fixes the mdEditor error mentioned in https://github.com/arj03/ssb-browser-demo/pull/170#issuecomment-773267088

Also fixes an issue where sometimes clicking suggestions did not work because the state change event would fire before the click and would hide the suggestions before the click could be processed.


arj03:
Thanks :)

PR #182, opened 2021-02-05
When blocking or unblocking, update the Profile page

### LINKED ISSUE #179: Refresh Profile when a user is blocked/unblocked
If you click "Unblock" on a profile, it does not rerender the page, so it still looks as though they are blocked.

KyleMaas:
Fixes #179.  Turns out it didn't need a full refresh.  Just the local version of blocking status needed to be updated.


arj03:
nice find

PR #188, opened 2021-02-06
Go to profile by name

### LINKED ISSUE #173: Go to profile by name in right corner
So we have this search thingy in the corner, would be nice if you could just do `@Kyle` and have suggestions come so you can go to a profile without known their id.

KyleMaas:
If you type an "@" in the upper right search box, it pops up a list of suggestions for profile names.  If you type in a profile name and hit enter or click one of the suggestions, it will take you to the corresponding profile.

Fixes #173.

### DIFF
--- a/browser.html
+++ b/browser.html
@@ -38,7 +38,20 @@
         <router-link to="/connections">{{ $t('pages.connections') }}</router-link><connected></connected> |
         <router-link to="/notifications" :title="$t('pages.notifications')">&#128276;</router-link> |
         <router-link to="/settings" :title="$t('pages.settings')">&#9881;</router-link>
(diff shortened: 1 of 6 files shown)


arj03:
This is working great. Really nice feature, could have used this yesterday when I was creating a ton of groups. The search box now floats a bit wierdly:

image

How does that look for you?

KyleMaas:
I had that periodically as well.  That row is just too cramped.  I had to wrap the box and suggestion popup in several layers of `<div>`s to get the positioning right, but that box should stay on the same line when there's enough space for it to do so.  We might need to shrink that box down some more.  Or have it expand and cover the menu when clicked.  (Thinking about it, this might be the better way to do it and would alleviate some of the crowding of that line.)

PR #194, opened 2021-02-06
Fix the Hops setting

### LINKED ISSUE #193: Test hops changes
Now that we have ssb-friends, we should test to see if the hops changes in Settings actually work.

KyleMaas:
This makes the Hops setting actually work.

Fixes #193.


KyleMaas:
Hm.  Well, this kind of works, but not all the way.  If I only follow myself and set Hops to 5, I still only get a graph containing 8 profiles.

arj03:
Yeah this probably also needs a bit of work in ssb-browser-core. To be honest it's not the feature I'm most excited about because I think 2 is a pretty good default. Anyway should probably be fixed.

KyleMaas:
Main use case for me is so that I can follow a bunch of people without growing my database too big.  See:

https://github.com/ssb-ngi-pointer/jitdb/issues/93

If I can set my hops to 1 or 2 (so it's only the people I follow), that would mostly work pretty well to keep the database small.  But if instead I set up ssb-browser-demo with a profile that only follows my desktop, and I set up my following/blocking on my desktop the way I want, then a higher hop count might be more appropriate and still only replicate the things I'm actually interested in.

KyleMaas:
With this:

https://github.com/arj03/ssb-browser-core/pull/26

...this feature works.

KyleMaas:
Correction: it works up to 2 hops.  Past that, ssb-friends is not giving any additional results, so a maximum of two hops are being synced.

KyleMaas:
Nevermind.  Wasn't testing it properly.  If I only follow my own account, setting hops gives me the following counts for SSB.getGraph:

* 1 = `{ following: Array(1), extended: Array(0), blocking: Array(0) }`
* 2 = `{ following: Array(1), extended: Array(8), blocking: Array(0) }`
* 3 = `{ following: Array(1), extended: Array(1116), blocking: Array(0) }`
* 4 = `{ following: Array(1), extended: Array(8121), blocking: Array(0) }`
* 5 = `{ following: Array(1), extended: Array(13716), blocking: Array(0) }`

With a setting of 4 or higher slowing the app down to a crawl as it tries to sync all that.

KyleMaas:
Oh, yeah, this most definitely works.  Hops set to 5 on the Between Two Worlds pub makes the app nearly unusable from pulling in that much data.  And setting it to 1 only pulls in data from direct followed profiles.

I didn't think about this before, but we probably should put a warning on the higher hop counts.  No way would I ever want to actually run with them, but someone else might if they're severely far off-grid and only able to access the network periodically.

KyleMaas:
There.  Now we have a warning.

Good way to stress test the app, though!

arj03:
Yeah, I'm wondering if we should limit it to say 3. 5 is pretty high. Otherwise this looks good.

KyleMaas:
Manyverse has:

* 1
* 2
* 3
* 4
* Unlimited

If you feel strongly about it, we can remove them, but I guess I'd rather give the user the option and warn them instead.

arj03:
I guess a warning is fine :)

arj03:
I plan on releasing what is in master now today. Just FYI :)

KyleMaas:
Sounds good.  I like the idea of releasing a bit more often, especially with all the bug fixes since the last release.

PR #199, opened 2021-02-06
Load as many blobs as possible when offline

### LINKED ISSUE #76: Make blob display independent of being connected to a peer
Markdown display of blobs is currently tied to needing to be connected to a peer (see https://github.com/arj03/ssb-browser-demo/issues/72#issuecomment-762553625 ).  I've worked around this with #75, but this also means that blobs can't be displayed at all if you're offline or if you're not currently connected to a peer with data (such as if you're connected to a room, and all other room-peer connections are offline).  It would be nicer if we had a way to try to display blobs if we already have them in cache.

KyleMaas:
Requires this to properly function:

https://github.com/arj03/ssb-browser-core/pull/24

This change tries to display as many blobs as possible when not connected to a server (such as when Use Offline mode is active).  It also contains a workaround for profile images trying to load before the profile index is filled in.

Fixes #76.

### DIFF
--- a/ui/markdown.js
+++ b/ui/markdown.js
@@ -4,14 +4,20 @@ const ref = require('ssb-ref')
 
 const mdOpts = {
   toUrl: (id) => {
-    var link = ref.parseLink(id)
-    if (link && ref.isBlob(link.link)) {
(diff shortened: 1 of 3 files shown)


arj03:
Looks good. Thanks, will test.

Issue #202, opened 2021-02-06
Mobile feedback


arj03:
Been testing the latest version on mobile. Fixed the nav bar and a minor thing in onboarding dialog. The profile editor needs a hand. It seems to need to go below the image if the width is not large enough. Can you help with that @KyleMaas?

arj03:
I merged in #203 but profile text editor still doesn't seem to work. Had to change the link it was broken before. I actually liked the small icon but couldn't fix it so changed it to You that works ;)

KyleMaas:
What kind of device are you testing this on?

arj03:
Sorry should have said, a Sony Xperia Xa2. Seems to be 1080p res

KyleMaas:
That's not going to be something I can duplicate exactly - don't have one of those.  But I can try to get close.  Which browser?

arj03:
Using Chrome.

KyleMaas:
I can't seem to duplicate this.  The profile description Markdown editor works just fine here on Android Chrome.  What exactly is it doing now?  Dropping down below the profile image?  Not displaying at all?

KyleMaas:
Well, I made a few changes (see #206), but I still can't seem to duplicate any issues with the Markdown editor with Chrome on an Android device.

KyleMaas:
Simulating every device I can in desktop Chromium, I do run into a few issues with:

* LG Optimus L70 (384x640)
* Nokia Lumia 520 (320x533)
* JioPhone 2 (240x320)

Various others are too narrow to display the editing buttons of the editor, but the editor still functions.  I'll try tweaking it for those and see if it helps you with yours.

KyleMaas:
#206 now has those three working.  Give that a try, please, and see if that fixes it for you.

PR #203, opened 2021-02-06
Mobile tweaks

KyleMaas:
See how this works.  It changes the size of the avatar images on Profile and rearranges the bottom menu (I didn't even know that was being used).

See #202.


arj03:
Oh my, I forgot to push up an update yesterday. Will fix the merge conflict.

PR #204, opened 2021-02-07
Group tweaks

### LINKED ISSUE #200: Group tweaks
As I alluded to earlier I absolutely loooooooooove groups. So I'm a really heavy user of that feature now. Some things I noticed:
 - Would be good with a loaded indicator
 - Would be nice if you could rename groups
 - Would be nice in the groups view that it shows the latest message, maybe only with time + author for each group
 - The add group should probably be on top

KyleMaas:
Implement suggestions from #200.

Fixes #200.

### DIFF
--- a/css/groups.css
+++ b/css/groups.css
@@ -28,3 +28,18 @@
   clear: both;
 }
 
+#groups .group
+{
+  margin-bottom: 4em;
+}
+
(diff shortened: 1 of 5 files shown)


arj03:
This is great. Thank you so much.

PR #206, opened 2021-02-07
Mobile Tweaks 2, Electric Boogaloo

KyleMaas:
This fixes the profile link at the bottom.  Turns out it was due to this:

[Prop Casing (camelCase vs kebab-case)](https://vuejs.org/v2/guide/components-props.html#Prop-Casing-camelCase-vs-kebab-case)

It also makes it so that it updates when you save your profile.

Still can't find a way to duplicate the Profile editor not working, though.

### DIFF
--- a/browser.html
+++ b/browser.html
@@ -62,7 +62,7 @@
         <router-link to="/private" class="navButton">{{ $t('pages.private') }}</router-link><new-private-messages></new-private-messages>
         <router-link to="/channels" class="navButton">{{ $t('pages.channels') }}</router-link>
         <router-link to="/groups" class="navButton">{{ $t('pages.groups') }}</router-link>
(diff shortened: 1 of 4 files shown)


KyleMaas:
Latest CSS tweaks fix the problems with these under a simulator:

* LG Optimus L70 (384x640)
* Nokia Lumia 520 (320x533)
* JioPhone 2 (240x320)

KyleMaas:
I also changed it so that if the screen width is too small for the lower nav menu to actually be usable, it switches back to traditional upper navigation.  It may take up more space, but it actually works, as opposed to when the lower nav menu is too wide and clips to the screen because it can't wrap.

arj03:
Tested this. The profile editor seems to be working now. Great fix on the profile link :)

KyleMaas:
Perfect.  Hard to debug that without being able to duplicate it.  Glad it worked.

PR #207, opened 2021-02-07
Show who's following you or other people

### LINKED ISSUE #187: See who's following a profile
ssb-friends supports doing `reverse` searches.  I think we could use that to see who's following a profile.

KyleMaas:
Shows a count of following/blocking, condenses some redundant CSS, and adds a section showing who's following you (or another profile).

Fixes #187.


arj03:
I like this functionality. Maybe we can remove the underline from the Following (x)?

PR #209, opened 2021-02-07
Convert upper Connections and Profile tabs to icons

### LINKED ISSUE #92: Connections under Settings?
Would it make sense to put Connections as a submenu under Settings?

KyleMaas:
Try and make a bit more space in the upper menu by converting Profile and Connections to icons like we did for the lower menu.  Also move the indicators so they don't take up so much space.

Fixes #92.

### DIFF
--- a/ui/profile.js
+++ b/ui/profile.js
@@ -289,9 +289,10 @@ module.exports = function () {
           if (err) return alert(err)
 
           // Re-render the lower nav menu.
-          if (this.imageBlobId != '')
+          if (this.imageBlobId != '') {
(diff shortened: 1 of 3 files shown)


ui/profile.js
```diff
@@ -290,6 +290,7 @@ module.exports = function () {
 
           // Re-render the lower nav menu.
           if (this.imageBlobId != '')
+            this.$root.$refs["upperNavProfileLink"].renderProfile({ name: "You", imageURL: this.image })
```
arj03, line 293: this change looks strange. Does it need a { } or is this a bug?

KyleMaas:
No, good spotting.  Fixed, and fixed merge conflicts.

arj03:
Great :)

PR #211, opened 2021-02-07
Remove underline from follow status lines

KyleMaas:
See https://github.com/arj03/ssb-browser-demo/pull/207#issuecomment-774744610


arj03:
Thanks :)

PR #215, opened 2021-02-08
Ask if the user wants to follow room endpoints

### LINKED ISSUE #64: Connect and follow
For Suggested Connections, it would be nice if you could connect and follow in one step instead of having to:

1. Click Connect.
2. After the connection is formed, navigate to the profile via the Connections link.
3. Click Follow.

KyleMaas:
When connecting to a room-endpoint, ask the user if they also want to follow the profile at the other end.

Fixes #64.


arj03:
Sound reasonable

PR #218, opened 2021-02-10
Groups filter for Public

KyleMaas:
One more step toward completing #32.  This ended up being more complicated than I expected due to group queries (including membership queries) being asynchronous.  This required the message queries to become asynchronous as well.


arj03:
Wonderful. Thank you

PR #220, opened 2021-02-10
Use FlexSearch instead of MiniSearch

KyleMaas:
See comments on https://github.com/arj03/ssb-browser-demo/pull/208#issuecomment-776764150

When I build that search function, I tried to make it as engine-agnostic as I could.  I originally chose MiniSearch because of its size.  This changes it to the compact version of FlexSearch.

@arj03 Does this work any better for you?  If not, it's not a big deal - it doesn't take much to swap out the engine.


arj03:
I found the problem. Take a look a my latest commit in master. Search is not really smooth :-)

Issue #219, opened 2021-02-10
Shrink main dist file


KyleMaas:
Main dist file is currently about 6.9MB (see https://github.com/arj03/ssb-browser-demo/pull/208#issuecomment-776765786).  Let's see if we can shrink that.

KyleMaas:
@arj03 Not finding a whole lot more opportunities to shrink this.  With #221 merged in, do you think we could close this?

arj03:
I have a tool I'd like to use to figure out what is taking this space. So lets keep it open for now.

KyleMaas:
This would help a little:

https://github.com/dweinstein/pull-async-filter/issues/1

arj03:
Now I remember the tool: `npm i -g disc`

# discify dist/js/bundle-ui.js > out.html

![image](https://user-images.githubusercontent.com/350682/107968405-9e8d8700-6fae-11eb-97b1-23a879e5206b.png)

Some of the biggest:

Toast: 1.2mb
Highlight: 1mb
libsodium: 500kb
codemirror: 400kb

It should be possible to disable highlight somehow I think. The gain we get from that package seems rather minimal.

arj03:
Yeah highlight.js comes from ssb-markdown. Maybe add a switch so it can be disabled?

arj03:
Bip39 is 200kb. Wonder where that is used.

KyleMaas:
I checked on Bip39 before.  It's used for mnemonic keys.

KyleMaas:
That's part of where I focused my efforts for minifying before, because I noticed the mnemonic key word lists have a *ton* of whitespace in them.  Part of the reason I've been working on shrinking the release bundle rather than the build bundle.

KyleMaas:
Well, that takes us down to about 3.5MB.  Do you think that's good enough, or is there more you'd like to try to do?

arj03:
I'll close this now. 1mb off compared to last release is pretty good. Thanks for the help with this @KyleMaas :)

KyleMaas:
You're welcome!

PR #221, opened 2021-02-10
Shrink size a little

KyleMaas:
This gets rid of a few unused dependencies.  Doesn't seem to actually do much to the output file size in debug builds, but probably good for housekeeping.

Takes the release build from 4520557 bytes to 4518019 bytes.

See #219.


KyleMaas:
Added terser to minify the output.  Release build size is now 4369664 bytes.  Still seems to work just fine.

* Tried terser with mangling.  Shrinks the size significantly, but does not work.
* Tried terser with only one compress pass.  Two passes does actually make the output smaller.

package.json
```diff
@@ -24,7 +22,6 @@
     "ssb-markdown": "^6.0.7",
     "ssb-mentions": "^0.5.2",
     "ssb-ref": "^2.13.3",
-    "ssb-sort": "^1.1.3",
```
arj03, line 27: this is used in ui/threads.js

arj03:
It's very nice to remove these unused stuff.

package.json
KyleMaas: Ah, you're right.

arj03:
Nice thanks :)

KyleMaas:
Unfortunately, it's not much.  But it's also not nothing.

PR #226, opened 2021-02-12
Update dependencies

KyleMaas:
Updates a few of the dependencies, drops an unsupported option from WorkBox, and moves WorkBox from being a regular dependency to only being a dev dependency.

Goes hand-in-hand with https://github.com/arj03/ssb-browser-core/pull/27

This does not contain all of the dependency updates from that, however.  This only contains the updates for this package.

It does seem to work with both of them updated.  When I first refreshed I got a buffer error that I sometimes get when refreshing, although I think that was likely unrelated since I sometimes get that anyway when refreshing a lot for testing.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -18,62 +18,71 @@
         "rimraf": "^3.0.2",
         "ssb-browser-core": "^8.2.0",
         "ssb-contact-msg": "^1.1.0",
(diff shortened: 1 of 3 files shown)


arj03:
Ok, will try

arj03:
This seems to have broken my local build. It can't find ssb-blob-files anymore. Have to look at this another time.

PR #227, opened 2021-02-13
Rebuild npm-shrinkwrap to fix the build process

KyleMaas:
> This seems to have broken my local build. It can't find ssb-blob-files anymore. Have to look at this another time.

From @arj03 at https://github.com/arj03/ssb-browser-demo/pull/226#issuecomment-778586267

I was able to duplicate this when not using my local ssb-browser-core branch I had updated dependencies on.  Since it was a dependency issue, I rebuilt npm-shrinkwrap.json with the dependency updates, and it fixed the build process.  It also eliminated the warning about package vulnerabilities.  And, in test running it, it seems to work fine with the dependency updates as far as I can tell.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -85,14 +85,20 @@
         "url": "https://opencollective.com/babel"
       }
     },
-    "node_modules/@babel/core/node_modules/source-map": {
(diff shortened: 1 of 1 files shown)


arj03:
That did the trick. Thanks :)

PR #230, opened 2021-02-13
Show profile changes

### LINKED ISSUE #229: Show something more than "undefined" for profile changes
From time to time, I've run into messages that just show up as "undefined".  Figured out why.  They're profile change messages that people replied to.  For example, see this message: `%WNA+DwuGi2iRwkKpJdxdIaKecxQyZ1rhpM9w5TYlp6w=.sha256`

We should have a good way to display those.

(Currently working on this one, but it's taking a bit, so I figured I'd better file an issue to keep from duplicating efforts in case of "great minds think alike".)

KyleMaas:
When someone has updated their profile, show what they changed.

Fixes #229.

### DIFF
--- a/ui/ssb-msg.js
+++ b/ui/ssb-msg.js
@@ -173,12 +173,12 @@ Vue.component('ssb-msg', {
 
     switch(this.msg.value.content.type) {
       case "about": {
-        var newInfo = "<p>" + this.$root.$t('common.profileUpdate', { user: (self.name || this.$root.$t('common.genericUsername')) }) + "</p>"
+        var newInfo = "<p>" + this.$root.$t((this.msg.value.author != this.msg.value.content.about ? 'common.profileUpdateOther' : 'common.profileUpdateSelf'), { user: (self.name || this.$root.$t('common.genericUsername')) }) + "</p>"
         self.body = newInfo
 
         // Try to find the next-to-latest profile update so we can show what changed.
         SSB.db.query(
-          and(author(this.msg.value.author), type('about')),
+          and(author(this.msg.value.content.about), type('about')),
           descending(),
           toCallback((err, msgs) => {
             var foundOurMsg = false
--- a/ui/ssb-msg.js
+++ b/ui/ssb-msg.js
@@ -8,7 +8,7 @@ Vue.component('ssb-msg', {
         <div class='header'>
           <span class="profile">
             <ssb-profile-link v-bind:key="msg.value.author" v-bind:feedId="msg.value.author"></ssb-profile-link>
-           </span>
+          </span>
           <span class='text'>
             <div class='date' :title='date'>{{ humandate }}</div>
             <router-link :to="{name: 'profile', params: { feedId: msg.value.author }}">{{ name }}</router-link> posted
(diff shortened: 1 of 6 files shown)


arj03:
I think this needs to test if its a self update or not. Or least it should show what you are updating. Also this seems to display very old entries? Like try going to my main profile. Anyway minor things, looking good otherwise :)

KyleMaas:
Self update works just fine here.  It displays it as "You posted a profile update".  Maybe I'm not understanding the issues you're having.

I'm not seeing any issues looking at your main profile.  I actually don't see any profile changes in the recent messages area for you, only follow changes (I'm running the branch that has both this and the follow/unfollow stuff).  What are you seeing?

KyleMaas:
I have noticed some weirdness with db2 lately.  Like for certain queries I'm getting duplicates of results.  But I haven't been able to nail down a good test case or way to duplicate it, so I haven't reported it because I don't know if it's actually in db2 or if somehow replication is stuffing duplicates into the database.  Perhaps that's related to what you're seeing?

arj03:
I think I know what the cause of the duplicates are.

arj03:
What i mean with self update is that if you do an about of something other than yourself it is also shown. Maybe it should only be the ones you assign to your own feed. We don't show the others anyway.

KyleMaas:
Ah, okay.  I'll see what I can do.

KyleMaas:
Something like that?

KyleMaas:
As long as you're around, is there anything else about this that needs to change?

arj03:
I was just about to look at this one

arj03:
I'm considering only showing this on profile page. What do you think?

KyleMaas:
I put it onto the Public tab as well due to Queen of England changing names and profile pictures and confusing the heck out of everyone.  Figured it'd be good to show that kind of information for all users, not just the one specific user you've already selected, because otherwise that kind of a change wouldn't become obvious.

ui/ssb-msg.js
```diff
@@ -171,7 +171,96 @@ Vue.component('ssb-msg', {
...
 
-    self.body = md.markdown(this.msg.value.content.text)
+    switch(this.msg.value.content.type) {
+      case "about": {
+        var newInfo = "<p>" + this.$root.$t((this.msg.value.author != this.msg.value.content.about ? 'common.profileUpdateOther' : 'common.profileUpdateSelf'), { user: (self.name || this.$root.$t('common.genericUsername')) }) + "</p>"
+        self.body = newInfo
+
+        // Try to find the next-to-latest profile update so we can show what changed.
+        SSB.db.query(
+          and(author(this.msg.value.content.about), type('about')),
```
arj03, line 181: This is not entirely correct as you might have other peoples about messages in here. Meaning author != content.about. Since we only show self-assigned abouts. Meaning we only show that kind of about changes.

arj03:
Good point. Lets leave in the public.

ui/ssb-msg.js
arj03: That was't really understandable ;) Since we only show show self-assigned abouts, I don't think we should display other kind of about changes as it's just confusing.

KyleMaas: How about that?

arj03:
Thanks :)

PR #231, opened 2021-02-14
Show follow/unfollow in Profile and Notifications

### LINKED ISSUE #225: Show follow / block on profile page
The same way we would add them to notfications. Not sure about public feed yet, at least there it would be to be grouped together, so might be good to start with them on profile page I think.

KyleMaas:
* Show when a profile follows/unfollows under the Profile page
* Show when other people follow/unfollow you under Notifications.

Based on #230.

Fixes #225.


arj03:
Seems to be working. Great work, really excited about this one! There is one thing I want to test on my profile that looks a bit wierd, unrelated to this PR. Anyway I had a comment on the other PR that would need to be fixed first.

arj03:
Can you fix the merge conflicts here @KyleMaas? Will review this now

arj03:
Nice, will be testing this now. Looks good. Thank you!

PR #233, opened 2021-02-15
Log connection errors

KyleMaas:
While trying to track down an issue, I ran into a problem where I was getting connection errors but couldn't trace why.  This makes it so that connection errors are logged.


arj03:
Thanks. I'll get to the other PRs as well. Been quite busy recently.

KyleMaas:
No problem.  Just trying to get as much done as I can before I start getting busy with work again and can't work on this as much.  Last few weeks have been quiet.  That is unusual for me.

PR #236, opened 2021-02-15
Patch out Highlight.js

KyleMaas:
More work on #219.  Patches out Highlight.js.  Takes release bundle down to 3558556 bytes - savings of 799711 bytes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "workbox-build": "^6.1.0"
   },
   "scripts": {
-    "postinstall": "pushd node_modules/pull-async-filter; sed -i 's/2.26.0/3.6.14/' package.json; rm -rf node_modules/pull-stream; popd; pushd node_modules/pull-goodbye; sed -i 's/3.5.0/3.6.14/' package.json; rm -rf node_modules/pull-stream; popd",
+    "postinstall": "pushd node_modules/pull-async-filter; sed -i 's/2.26.0/3.6.14/' package.json; rm -rf node_modules/pull-stream; popd; pushd node_modules/pull-goodbye; sed -i 's/3.5.0/3.6.14/' package.json; rm -rf node_modules/pull-stream; popd; pushd node_modules/ssb-markdown; sed -i -e '/highlight: function/,+8 d' -e '/highlight/d' lib/block.js; sed -i '/highlight.js/d' package.json; popd; rm -rf node_modules/highlight.js",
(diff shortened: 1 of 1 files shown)


arj03:
Sweet.

KyleMaas:
I got thinking about this - since this is patching a source file and removing specifically 9 lines, we might want to lock ssb-markdown's version in place and only upgrade when we've tested it, rather than doing a hat dependency where the file might change out from under us.

arj03:
Yeah but we do have the shrinkmap file. This postinstall is pnpm specific, right?

KyleMaas:
npm-specific, yes.

arj03:
Wierd, its not working here. Are you using bash shell?

KyleMaas:
Yes.  Is it giving you an error?

arj03:
Yeah:

```
> pushd node_modules/pull-async-filter; sed -i 's/2.26.0/3.6.14/' package.json; rm -rf node_modules/pull-stream; popd; pushd node_modules/pull-goodbye; sed -i 's/3.5.0/3.6.14/' package.json; rm -rf node_modules/pull-stream; popd; pushd node_modules/ssb-markdown; sed -i -e '/highlight: function/,+8 d' -e '/highlight/d' lib/block.js; sed -i '/highlight.js/d' package.json; popd; rm -rf node_modules/highlight.js

sh: 1: pushd: not found
sh: 1: popd: not found
sh: 1: pushd: not found
sh: 1: popd: not found
sh: 1: pushd: not found
sed: can't read lib/block.js: No such file or directory
sh: 1: popd: not found
```

KyleMaas:
Oh, you're using `sh`, not `bash`.

arj03:
npm must be. I should be running bash.

arj03:
`npm config set script-shell bash `worked, but I would rather not have to instruct users to do that.

KyleMaas:
Well, I suppose if you need to, those can be changed to `cd nodule_modules/pull-async-filter` (etc.) and `cd ../..`  I just prefer pushd and popd because I feel they're a little more robust in case of failures.

arj03:
For browser-core we have used the patches folder. That always works. Good old `patch` :-)

KyleMaas:
How are patches in that folder applied?  I don't see anything in ssb-browser-core that actually does that.

arj03:
It's manually applied as it's optional. Because it just decreases the bundle size. Like these ones.

PR #239, opened 2021-02-15
Eliminate pushd and popd from cleanup script

### LINKED ISSUE #238: npm i requires bash configured
`npm config set script-shell bash `worked, but I would rather not have to instruct users to do that.

_Originally posted by @arj03 in https://github.com/arj03/ssb-browser-demo/issues/236#issuecomment-779424058_

KyleMaas:
Fixes #238.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "workbox-build": "^6.1.0"
   },
   "scripts": {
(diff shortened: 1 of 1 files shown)


arj03:
:+1:

PR #242, opened 2021-02-16
Fix missing dependency on ssb-keys

KyleMaas:
Turns out we were using ssb-keys, but only as a dependency of other stuff.  In the case of it not being able to be deduplicated, npm could end up installing it only under the other packages requiring it, which caused a dependency error for ssb-browser-demo.  This fixes that by pulling it in directly.


arj03:
indeed, thx

PR #244, opened 2021-02-16
Dedupe more

KyleMaas:
Takes release bundle from 3559975 bytes to 3514373 bytes - a savings of 45602 bytes.  Diminishing returns, but it's still something.  This also splits out the dedupe patching into a shell script which runs postinstall instead of trying to inline it all.


arj03:
This feels a lot cleaner way to patch things. Nice!

PR #246, opened 2021-02-18
Add last updated information to connections

### LINKED ISSUE #141: Per-peer "Last synced"
It would be nice if Connections displayed information about when a peer was last synced with.  It would be a decent way of finding out if you have any dead peers which haven't been up in a long time.  It would also give you an idea of how recent your copy of messages is.

KyleMaas:
Adds information to the connections list to show when a connection last received an update.

Interestingly, this also allows you to see how long it takes for the receive queue to flush to the database and be indexed, since if you pull a large batch of data (follow someone with a lot of followers you don't already replicate), you can see the last updated change quickly and then watch the profiles fill in on the progress bar as they complete indexing.

Fixes #141.

### DIFF
--- a/messages.json
+++ b/messages.json
@@ -89,6 +89,7 @@
       "existingConnections": "Connections",
       "unsupportedConnectionTypeX": "This SSB client does not support connections of type '{connType}'.\n\nUnfortunately, browsers place a lot of restrictions on how we can connect.  Most communications have to be done over WebSockets, and the invite code you're using doesn't appear to support them.  Sorry!",
       "followNewConnection": "Do you want to also follow this person so they autoconnect and sync in the future?",
+      "lastUpdateReceived": "(Last update received {datetime})",
       "dbStatus": "DB status",
       "ebtStatus": "EBT status"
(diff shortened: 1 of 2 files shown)


KyleMaas:
I should note that this *should* work without the latest change to ssb-browser-core, but for better results, it should be used with a version of ssb-browser-core which includes https://github.com/arj03/ssb-browser-core/pull/30

arj03:
One question, does the timestamp also update even if no data is coming through, I'm guessing the default ping/pong will update this ts as well?

KyleMaas:
The `ts` value from EBT is only updated when a message is received.  I have no idea where the peer `hubUpdated` and `stateChange` timestamps come from.  But that's why I set the text as "Last update received" rather than "Last synced" or "Last talked to the server".  It's a reflection of when we actually received updated information from that peer.

arj03:
Thanks, created a new version of core that I will be pushing up now.

KyleMaas:
Perfect.  Thanks!

PR #247, opened 2021-02-19
Add support for blog-type messages

### LINKED ISSUE #223: Support blog
At least show them. It's actually relatively easy because they are just blobs. There is a library [here](https://github.com/ssbc/scuttle-blog). Like the SSBC newsletters is published as a blog.

KyleMaas:
Adds support for "blog" messages.  Theoretically, they should be UTF-16, but in reality when I tried decoding existing messages as UTF-16, it came out as garbage.  So this decodes them as UTF-8, which is how most of the messages I found were actually encoded.

Fixes #223.

### DIFF
--- a/ui/ssb-msg.js
+++ b/ui/ssb-msg.js
@@ -1,6 +1,7 @@
 const human = require('human-time')
 const md = require('./markdown')
 const helpers = require('./helpers')
+const pull = require('pull-stream')
 
 Vue.component('ssb-msg', {
   template: `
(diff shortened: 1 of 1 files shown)


arj03:
Was looking at this 2 times because I didn't think it was this easy. Nice one :-)

KyleMaas:
Yeah.  I think it wouldn't be a bad idea, at some point, to add the ability to let users choose to post as a blog post instead if their post is too long rather than just erroring out.  But that's a pretty low priority.  Probably ought to file an issue for that so I don't forget, though.

PR #248, opened 2021-02-19
Fix profile search

KyleMaas:
Evidently the profile search (type "@" and then something in the search bar, hit Enter) was broken because of the transition from the profiles index to aboutSelf.  This fixes that.


arj03:
Great find.

PR #250, opened 2021-02-19
Retry loading profile after refresh

### LINKED ISSUE #51: Refreshing while looking at a profile doesn't load the profile properly
If you're looking at a profile and you refresh the page, it does not show the details of the profile after the page is loaded.  I suspect this is because when the page is loaded, it's not connected yet, but I have not yet traced this down.

KyleMaas:
If you refresh the page while looking at a profile, previously it would attempt to load the profile immediately before the index loaded, resulting in an empty result.  This changes that so that if the index isn't loaded, it waits several seconds to let the index load and then tries again.  If that doesn't work, it waits for a connection with data to load the profile.

Fixes #51.

### DIFF
--- a/ui/profile.js
+++ b/ui/profile.js
@@ -547,50 +547,67 @@ module.exports = function () {
         this.updateFollowers()
         this.updateBlockingUs()
 
-        const profile = SSB.getProfile(this.feedId)
-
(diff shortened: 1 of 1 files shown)


arj03:
thanks

Issue #234, opened 2021-02-15
invalid message: expected sequence 2 but got:1


KyleMaas:
Figured I might as well file an issue for this so we can collect together information about it.  I'm getting this fairly consistently now in a lot of different situations and I don't understand enough about what it's doing to be able to track it down.

Steps to reproduce:
1. Start a new profile (open it in a private browser window or something).
2. Leave everything in the onboarding screen as defaults and click Get Started.

That's all it takes for me to trigger that error.

KyleMaas:
Also, can you give me any pointers as to where to start debugging this?

arj03:
The error is from [validate](https://github.com/ssb-js/ssb-validate). So either [here](https://github.com/ssb-ngi-pointer/ssb-db2/blob/b7522decb7c87803791d155c574a38956ceb5571/db.js#L151) or [here](https://github.com/ssb-ngi-pointer/ssb-db2/blob/b7522decb7c87803791d155c574a38956ceb5571/db.js#L180). It must be somehow related to how ssb-browser-core does the partial replication.

KyleMaas:
How long do you think until https://github.com/ssb-ngi-pointer/ssb-subset-replication is usable?

arj03:
We are aiming for May

KyleMaas:
@arj03 Probably could use your help on this one, too.  Having a hard time making any progress at all on this.

arj03:
Nice find. I can reproduce this problem.

arj03:
Fixed with this [commit](https://github.com/arj03/ssb-browser-core/commit/c9268d63ee0eb77f82e7e85225aae57628fc48ea)

KyleMaas:
Hokey smokes!  That's all it took?  The difference it makes to have the knowledge of how it all works...

I'll give it a test here in a few minutes.

arj03:
Yeah, a bit of a funny bug. Very nice to have it fixed though, might have caused other kinds of wierd problems.

arj03:
I'll close this. I made a release with these changes :-)

KyleMaas:
Sounds good.  Just finished doing a write-up on one-way follows.

Yep, looks like that fixed it.  Thanks!

arj03:
Sweet! Thanks for reporting this.

Issue #186, opened 2021-02-06
One-way follow in a room prevents proper sync


KyleMaas:
Steps to duplicate:

1. Open ssb-browser-demo in two different browser windows with no prior state (either private tabs or cleared profile, or on two different machines.  Do not connect to a pub, but join the Between Two Worlds room on both.
2. In Connections on one of the machines, connect to the Possible Connection for the other one.
3. On one of the browsers (we'll call it Browser A), click the other browser's profile under Connections and click Follow.
4. After a few seconds, Browser A will show up as fully synced.  Browser B does not show as synced according to its Connections indicator.
5. Post a test message to Public from Browser B.
6. Observe that the message does not show up in Browser A.  You can even refresh Browser A and you still won't get the message from Browser B.

That demonstrates the problem.  But then if you follow Browser A from within Browser B, the test message from Browser B is finally synced to Browser A.

Any idea why this would be happening?

KyleMaas:
I think this might be related to this:

https://github.com/ssbc/ssb-ebt/issues/33

KyleMaas:
Just tried this with just one browser connecting to the room and then connecting and following one of @arj03's accounts.  A few seconds after I do that, I get error messages like this:

```
no stream for incoming msg {req: -7, stream: true, end: true, value: {…}, length: 829, …}
```

With the JSON being something like this:

```
{
  end: true,
  length: 829,
  req: -7,
  stream: true,
  type: 2,
  value:
  message: "shs.client: error when expecting server to accept challenge (phase 1).↵possibly the server is busy, does not speak shs or uses a different application cap",
  name: "Error",
  stack: "Error: shs.client: error when expecting server to accept challenge (phase 1).↵possibly the server is busy, does not speak shs or uses a different application cap...(this is huge - not putting it all here because it's likely irrelevant)"
}
```

KyleMaas:
(If I remember correctly from the ID, I'm pretty sure that was "browser arj" I was trying to connect to before.  Since there was no connection, and I wasn't connected to a pub, it didn't have the profile name at the time.  And I forgot to take note of it.)

KyleMaas:
I did not get the same error when I tried connecting between two browsers on my own computer.  So maybe that was a fluke.

KyleMaas:
I do get this error, though (which is the same one I get when I first sync the Between Two Worlds pub on a new profile):

```
[...]

PR #252, opened 2021-02-19
Add new Threads tab which displays a forum view of public posts

### LINKED ISSUE #228: Forum view
> In fact, it’s the first time I use a phpBB in a looooooong time. The UI is… amazing!!!
It has no hidden things, no things that move around based on some open-source-yet-totally-unintuitive algo, it has virtually no client-side JS, it’s got strong, tree-like categories.
> 
> I keep saying it and it bears repeating: ssb needs a UI like this, even if it is just a heuristic view onto the existing posts.

From Daan in `%M2LuCm+C9piUuZDCpKk3uvqWqIprB%2FTisVh7B0+oRns=.sha256`

As a still-active user of old-style fora, I would have to agree with this.  It would make it much easier to browse posts.  And I think it's something that could be implemented without too much difficulty.

KyleMaas:
Adds a forum view of Public posts.

Fixes #228.


KyleMaas:
I'll note that for the moment, this just displays public posts in a forum-ish view.  But it could easily be extended to show the latest messages in favorited channels or messages from user groups.  But I wanted to start with this and make sure I was going in the right direction first.  I actually find this far more useful than the "drinking from a firehose" Public tab.

arj03:
Got a bunch of:

```
this.created is not a function
    at a.feedId (bundle-ui.js:149536)
    at fn.run (vue.min.js:6)
    at un (vue.min.js:6)
    at Array.<anonymous> (vue.min.js:6)
```

arj03:
And yes this is indeed a quite nice condensed view. A suggestion would be to add a miniature picture of the author to the right so it spans both the name and the time. Also there seems to be something with Last post. It's always the same as started by. Havn't looked at the code yet. 

And yes I like this view as well.

arj03:
Actually not sure about the image now. One could also add one for the original author to the left, but there is a certain asthatic to just the raw text format. Just bouncing ideas here.

KyleMaas:
The name on the right always being the same as the thread starter would be related to the this.created error.  It worked here, but looking at it, I could see how that could crash.

And, yeah, I thought about doing the profile picture and then figured it might be better with just the text.  It's a little more condensed and forum-y this way, IMHO.

arj03:
That fixed the issue

KyleMaas:
That should fix the error.

I also added tiny little profile images next to the names.  That doesn't expand it much.  How's that look?

arj03:
Yeah the tiny images are great I think. Still has the old school forum vibe but a lot easier to scan.

KyleMaas:
Agreed.  It's a lot easier to identify users that way.  Good call.

KyleMaas:
Enhancements I could see adding to this:

* Of course, showing things other than just Public messages, such as channels and user groups.
* Proper pagination.
* I'd like to port this (and the Thread page) to ssb-threads so both are consistent.
* For a faster query, this pulls thread roots sorted descending.  That's less helpful than if new messages floated a thread to the top of the list.  But the logic for determining that is a little bit more complicated, particularly when considering pagination.
* We should have a Refresh button and automatic updates like Public.
* It would be nice if this had a "New Thread" button which pops up a compose box (like Public).
* Would be nice if we could track which messages have been read so we could show threads with unread messages in a different color or something.
* Visually, this may need a few tweaks on the different themes.
* We probably ought to have a maximum width for the most recent post column so long names don't make that so wide that other columns get scrunched.

But, by the same token, I don't want this to turn in a behemoth of a monolith that creates conflicts every time we want to change something, so I'm thinking these would be better as individual pull requests.

KyleMaas:
I think I'd like to try and get this stuff merged before I work on the concurrency stuff, since I'm fairly certain it'll require the same modifications as everything else and I'd rather have that be a nice clean pull request.  What else needs to be done on this for it to be merged?

ui/threads.js
```diff
@@ -87,6 +87,24 @@ module.exports = function (componentsState) {
         }
       },
 
+      pullRoots: function(categoryKey, messages) {
+        var self = this
+        var rootKeys = messages.map((x) => { return x.value.content.root || x.key })
+          .filter((x, index, self) => { return self.indexOf(x) == index })
```
arj03, line 139: You can use a Set here instead of the filter (unique), it's faster and a bit easier to see what is going on.

arj03:
Just a minor thing I found. This seems to be something like 12 seconds on my laptop to display. There must be something we can do to speed this up. In a way this view seems very similar to what ssb-threads is doing. So maybe it makes sense to see if we can adapt that and use here?

arj03:
I can see why you'd want this merged first. This is also why I took a look and tried it out, but as it is now I would rather try and get some fixes in than putting in more features. I'm thinking of doing a release soon again.

ui/threads.js
KyleMaas: Haven't used Set before.  It's a little less readable to me, but take a look and see what you think.

KyleMaas:
> The comment was just a minor thing I found. This seems to be something like 12 seconds on my laptop to display. There must be something we can do to speed this up. In a way this view seems very similar to what ssb-threads is doing. So maybe it makes sense to see if we can adapt that and use here?

This is largely due to the second query.  For some reason, asking for a bunch of messages by or'd keys seems to be really slow.  It was much faster prior to commit `a917df1`.

ssb-threads was on my to-do list.  I've been watching @staltz's progress on porting to db2 for a while now.  But since that involves a change to ssb-browser-core and rebuilding the Thread page, I figured it'd be better for after we got this behaving the way we wanted.

KyleMaas:
> I can see why you'd want this merged first. This is also why I took a look and tried it out, but as it is now I would rather try and get some fixes in than putting in more features. I'm thinking of doing a release soon again.

Makes sense.  But I'd be of the opinion that that release should happen before any concurrency fixes anyway, since there will be a lot of changes that will need to be tested.  Really, I don't see any reason why a release couldn't happen now (or at least pretty soon).  It's been pretty stable for me lately, as evidenced by the lack of "bug" issues I've been reporting.  The two biggest and most serious bugs I'd love to see fixed are #234 and #186, both of which have proved to be outside of my competency level to debug.

KyleMaas:
Well, turns out Thread probably shouldn't be ported to ssb-threads because it uses partial-replication to try to fetch missing messages.  But it works pretty well for the forum-like view.

KyleMaas:
Oh, I should note that this relies on this now: https://github.com/arj03/ssb-browser-core/pull/33

KyleMaas:
Is this working any better for you?

KyleMaas:
There we go.  Merge conflicts fixed, and working again.

arj03:
Great, will have a look soon.

arj03:
Hey Kyle. I havn't forgotten about this. Have just been SUUUUPER busy. Hope you understand that there is a human in the other end :-) I will try to get to this soon.

KyleMaas:
Heheh...thanks for the update.  I was worried for a bit that this was just going to languish away.  I've at least been using it in the meantime.  :)

arj03:
Testing this now. A lot faster, nice to see ssb-threads used here.

I ran into a strange issue, the first thread showed andre's phone for a bit and then changed into this invalid date and no image:

image

arj03:
This is looking good code wise, we can always fix minor issues after merge.

KyleMaas:
Hm.  Not sure what would cause that.  It's never done anything like that for me before, but I'll see if I can figure it out.

KyleMaas:
Oh, and thanks for merging this!  I've actually been using the Threads tab a lot lately, and there were several things I was kind of holding off on working on until we got this done because I didn't want to have to deal with the merge conflicts it would cause.

arj03:
Yeah it does seem to work pretty well. A whole different overview of what is going on. I especially like the replies count.

PR #256, opened 2021-02-20
Add Reset Database button

### LINKED ISSUE #35: Feature request: Buttons to reset the database or scan for corruption
As many times as I've managed to corrupt my local database, it would be awfully nice if there was a button I could click to scan the database and fix problems and/or a button to reset the database to clear it of anything it has synced from anywhere else leaving only my local messages, since I suspect most of the corruption I'm experiencing is due to refreshing partway through a write to the database while syncing from the pub.

KyleMaas:
Theoretically, this would do it.  In reality, it does not work.  It clears the indexes and blobs AFAICT.  But it does not clear the database - `removeDB` just doesn't seem to work.  So I figured I'd file this so we'd have something to test with.

This depends on https://github.com/arj03/ssb-browser-core/issues/32

Fixes #35.  At least if we can get it working.


KyleMaas:
Can't realistically do this until https://github.com/ssb-ngi-pointer/ssb-db2/issues/198

Not seeing any progress on that front, I think I'm going to close this.

Issue #139, opened 2021-01-30
Concurrency issue - Posting from two different tabs windows does not work


KyleMaas:
If you open two different browser tabs (for example, to look at two different threads) and try to post to one, then switch to the other tab and post there, the messages from one will not show up on the other.  It seems to behave as if you've forked your identity.

KyleMaas:
I've done a little bit of investigation on this.

It looks like the only way to share an SSB object (which I think is going to be the only way to solve the concurrency issue) is to make sure that all "new tab" operations (center-clicking things) are converted into window.open() calls so that we can access window.opener.SSB from the child tab.  We will then need to keep a reference around to each of those child windows so that we can notify them if the parent window closes, because we will lose our SSB state and they will need to pause.  This would also require that all references to the global SSB object are instead redirected to parent windows if we have them.  It's doable, and indirection could be done synchronously, making it relatively easy to drop into place.  It's not exactly easy, but it's doable.

Now, the better solution, if we can swing it: Theoretically, it might also be possible for a parent window to pass each child window references to all other child windows so that if the parent window is closed it can elect one child window to be the holder of SSB.  After closure, that child window would have to re-initialize SSB (because I don't think we can just pass the reference, and it wouldn't have any of the timers it needs for connection scheduling and such) and then proceed.  This is arguably a more robust way to do it if we can pull it off, but it would require that all references to global SSB be converted into an asynchronous callback so that child windows would continue running without crashing while the newly-elected SSB holder reinitializes its SSB object.

But to know if that's even realistic, I'll have to do some more testing.

KyleMaas:
Looks like another option for getting around `window.open()` conversion might possibly be `rel="opener"` on all links, which is a lot more doable.  See:

https://developer.mozilla.org/en-US/docs/Web/API/Window/opener

Also appears to be supported in at least recent Firefox and Blink-based browsers, probably also WebKit-based.  Good possibility IE does not support it (Edge should).  More investigation needed.

KyleMaas:
`rel="opener"` is likely not going to work.  I missed the fact that it only works for links which are also `target="_blank"`.

To use `window.open()` instead would mean we'd need to hook the `click` event on all links so we could catch center clicks as well as `contextmenu` events so we could hook right-clicking to open a new tab and use `window.open()` instead.  Messy, but doable.

[...]

PR #259, opened 2021-02-23
Basic protection against concurrency issues

KyleMaas:
Basic protection against concurrency issues.  See #139.  Uses a localStorage-based mutex which times out in case a tab dies on us.  Adds a bunch of extra checks throughout the codebase to try to gracefully handle a locked SSB.

Also fixes:
* Fix error where clicking Refresh when you didn't have any messages would error out.
* Fix missing getProfileNameAsync in Private.

### DIFF
--- a/browser.html
+++ b/browser.html
@@ -28,6 +28,12 @@
     <div class="refresher">
       <img src="hermies.png">
     </div>
+    <div class="modalError">
+      <div>
+        <div id="modalErrorMessage">
+        </div>
+      </div>
+    </div>
     <div id="app">
       <h1>{{ appTitle }}</h1>
       <p id="navigation">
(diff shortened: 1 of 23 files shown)


arj03:
This is really exciting! I'll be in a call with mix soon so don't have time to properly review this, and it needs because it's a quite big change. Is it ok if I create a branch with some changes while I review it? Have some ideas to make this a bit simpler maybe. Also if you create your branches in this repo instead of KyleMass, then it should be easier to have that kind of workflow.

KyleMaas:
Go for it.  This should allow you to make changes, though - I left that on.

If you do make changes, there are a few important things about this:

1. It is possible for a not-quite-fully-initialized SSB to be passed back to the caller.  This pretty much always happens with the first getSSB() call that initializes the SSB object.  This is why in most places I don't just check for whether the SSB object is returned, but also that the functionality it needs is included in it.
2. For long-running processes (like Connections status polling) or anything with a callback, it is possible for the parent window to be closed between the time the callback was set up and when it runs.  So the SSB object needs to be acquired anew (and error checking needs to happen) each time one of these runs.
3. I know you wanted this done synchronously, but a lot of the places that need SSB could be vastly simplified with an asynchronous callback.  As it is, there are a lot of setTimeout()s to just try again.  I intentionally set this up so it was easy to convert over.
4. The mutex system is localStorage-based.  This is because it's very widely-supported across browsers, it uses the same browser API as the storage we're protecting in many cases, and per the localStorage spec browsers are required to implement mutexes in their implementations.  That was tricky to get right, but there is a reason it delays startup by 2 seconds, and that's to check for pings from other tabs.  Without that, it starts nearly immediately but does not properly protect against concurrent access.

KyleMaas:
Also, if you can commit early and often, that'd be great, because my next step is to work on getting center clicks and context menus programmatically opening child windows so they can be tied in.  I plan on doing the same.  Don't need any more conflicts than necessary.

KyleMaas:
Moved to #260.

PR #265, opened 2021-02-26
Message context menu

### LINKED ISSUE #263: Add ssb-msg context menu (right-click)
It would be awfully nice if `ssb-msg` had a right-click menu for the message which had "Copy ID" and "View Source".  That way, if you ran into an issue where a message was behaving strangely, you could examine it much easier than trying to query through the database and find it or step through the code and examine internal variables.

KyleMaas:
Adds a context menu (right-click) for ssb-msg so you can right-click a message and choose one of the following options:

* View whole thread
* Open thread in new tab
* Copy message ID
* Copy thread root ID
* View source

Fixes #263.

### DIFF
--- a/browser.html
+++ b/browser.html
@@ -35,7 +35,7 @@
       </div>
     </div>
     <div id="app">
-      <link-contextmenu ref="linkContextMenu"></link-contextmenu>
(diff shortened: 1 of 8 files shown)


arj03:
Looks good, thanks!

Issue #140, opened 2021-01-30
Sync button


KyleMaas:
From what I've seen of older screenshots, ssb-browser-demo used to have a sync button to manually sync right away rather than wait for the scheduler to kick in.  It would be nice if we had that feature back.

arj03:
I think this was before ssb conn. I would actually rather have an offline mode.

arj03:
I guess we can close this now?

KyleMaas:
Not sure.  I have still run into situations where the scheduler didn't actually kick in until after I refreshed the page.  But I haven't been able to track those down yet to see why.  I suppose you could do the same thing with going offline and then online again, though.

KyleMaas:
It just did it again on me.  Clicking the refresh button on Public did nothing.  But I refreshed the page, and I got about 2 hours' worth of messages that came in.  This, on the console, seems to also kick it out of a stuck position:

`SSB.net.connScheduler.stop();SSB.net.connScheduler.start()`

Doesn't even need a delay between them to work.  So I'm thinking if we did do a "Manual Sync" button, that's probably all it would need to do.

arj03:
Wierd. Havn't seen that. I would of course rather fix the sync issue than having a manual sync button.

KyleMaas:
Sync stopped again.  Connections screen said both followed and extended were 100% synced, but the latest Public message was about 4 hours old.  Tried going offline and online, and it never tried connecting to the pub again.  Refreshed the page.  Still didn't try to connect to the pub.  Manually clicked the button to connect to the pub.  Several seconds later, I had 49 new messages on the Public tab.

I noticed in the logs that there were several times the pub server's stream was closed before it finally quit.  Is there a limit to the number of connection retries before a connection is considered dead and is not retried again?

KyleMaas:
Got stuck again.  Last message was listed as several hours ago.  Refreshed the browser, it reconnected to the pub and downloaded recent messages.  Still have no idea why it keeps doing this.

KyleMaas:
Stuck again.  Last message listed as "12 hours ago".  After that is "12 hours ago" and "21 hours ago".  Connections shows "Last update received" on the pub as about 30 minutes ago.  Followed sync and Extended sync are both showing as 100% complete.  "Use Offline" and then "Connect Online" did not fix the problem.

DB status in stuck condition:

```
{
  "log": 178495373,
  "jit": {},
  "indexes": {
    "keys": 178495373,
[...]

PR #268, opened 2021-03-05
Add setting to enable detailed logging (EBT logging for now)

KyleMaas:
Add a setting to enable advanced detailed logging.  See:

https://github.com/arj03/ssb-browser-demo/issues/140#issuecomment-790889512


arj03:
Will have a look at this soon. 

On a related note I just fixed a bug in EBT that was merged today. I updated between-two-worlds with the fix which means that I get data from a few more feeds than before.

KyleMaas:
This look okay?

arj03:
Yeah thx

KyleMaas:
Thanks!

PR #277, opened 2021-03-20
Add thread tooltips

### LINKED ISSUE #275: Tooltip preview when hovering the mouse over a thread title
Considering how short thread titles are truncated to fit on the Threads page, it would be nice if you could hover your mouse over them and see the full title and a short preview of the first/last message in the thread.  This is something I keep instinctively doing that I'm used to from other forums, and I keep forgetting it doesn't work here yet.

KyleMaas:
Adds tooltips for the Threads page to show a preview of the first and the latest message in each thread.

Fixes #275.

### DIFF
--- a/ui/helpers.js
+++ b/ui/helpers.js
@@ -53,6 +53,15 @@ exports.getMessageTitle = function(msgId, msg) {
   return msgId
 }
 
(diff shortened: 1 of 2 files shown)


arj03:
Not using threads much, but can see how this might be quite useful.

PR #278, opened 2021-03-20
Make sure replying to a Thread shows feedback of the posting

### LINKED ISSUE #271: When replying to a thread, refresh the Thread page
When you reply to a thread from the Thread page, the reply does not always show up until you refresh the page.  We should definitely show the new message immediately.  (I'm thinking it would also be a good idea to check for new messages in that thread in case someone replied before you so you can see if your new message might somehow be taken out of context, so maybe a full refresh is in order.)

KyleMaas:
I was not able to duplicate the problem in #271, but looking at the code, it makes sense how it could happen.  This makes it so the user has immediate feedback of their posting once it's been added to the database.  It also fixes a few bugs relating to null message keys - there was already code there for that case, so I have to assume something else was able to trigger that before, but it wasn't fully working.

Fixes #271.

### DIFF
--- a/ui/ssb-msg.js
+++ b/ui/ssb-msg.js
@@ -27,7 +27,7 @@ Vue.component('ssb-msg', {
           </span>
         </div>
 
-        <h2 v-if="msg.value.content.subject || msg.value.content.title">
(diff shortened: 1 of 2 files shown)


arj03:
Thanks

PR #280, opened 2021-03-25
Fix minor SSB bug for really slow situations

### LINKED ISSUE #279: When browser's running slow, new-public-messages can error
new-public-messages calls the syncronous version of `getSSB()` and then checks for an error.  Unfortunately, this means that if the browser is running exceptionally slowly (such as in a resource-constrained environment), new-public-messages can end up trying to run `SSB.db.query()` on an `SSB` object which doesn't have a `db` yet.  The `err` check needs to also check for the presence of `SSB.db` and retry if it's not initialized yet.

KyleMaas:
Fixes #279.

### DIFF
--- a/ui/new-private-messages.js
+++ b/ui/new-private-messages.js
@@ -24,7 +24,7 @@ module.exports = function (state) {
         var self = this;
   
         [ err, SSB ] = ssbSingleton.getSSB()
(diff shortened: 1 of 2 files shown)


arj03:
Nice, thanks :)

PR #285, opened 2021-03-27
Add option to enable/disable DHT connections

KyleMaas:
Add a UI option to enable/disable DHT functionality.

Fixes https://github.com/arj03/ssb-browser-core/issues/47


arj03:
Great, I can see that it works by fixing connections. I'll leave a comment about that in ssb-browser-core. Thanks!

PR #287, opened 2021-04-06
Portuguese review

tdruiva:
I was comparing the Portuguese version with the English messages and here goes some suggestions, by the way I tried to keep gender-neutral.

cc @luandro

### DIFF
--- a/messages.json
+++ b/messages.json
@@ -523,7 +523,7 @@
     "channel": {
       "blankFieldError": "Por favor, forneça uma mensagem para publicar.",
       "postNewMessage": "Poste uma nova mensagem",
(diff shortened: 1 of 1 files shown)


arj03:
Thank you :) Naturally I have no chance to review this, so I'll wait for Luandro. I must say that I tried out the translations and found a few places where things could not be translated so I fixed a few of those. If you find more you can add an issue and we can get those fixed.

luandro:
Amazing! Thank you @tdruiva! Reviewed, and ready to merge.

arj03:
Thanks to both of you :-)

PR #297, opened 2021-06-28
Unbermuda integration

arj03:


KyleMaas:
Pubs seem to be working okay, but the room isn't working (guessing because the room server's down right now) and DHT sort of works but causes errors:

```
bundle-ui.js:165372 Uncaught TypeError: Cannot read property 'updatePartialState' of undefined
    at bundle-ui.js:165372
    at bundle-ui.js:118359
    at bundle-ui.js:118255
    at bundle-ui.js:118511
    at PacketStreamSubstream.weird.read (bundle-ui.js:107218)
    at PacketStream._onstream (bundle-ui.js:111830)
    at PacketStream.write (bundle-ui.js:111764)
    at bundle-ui.js:107241
    at bundle-ui.js:118259
    at bundle-ui.js:118565
```

```
Access to XMLHttpRequest at 'dht:kXhzkgY3eVFhS0X69J+eSG1bvD4NX6LEeveOEV1u82U=:@f/K1EUSJ7C4vyTfVpF6PQKkQfgUl9fTZyX5zPf7eITs=.ed25519/blobs/get/&cRXdDK2vjSgGIalEyyZJZto51dSXKi16yUdcQ5N7eGs=.sha256' from origin 'http://127.0.0.1:8080' has been blocked by CORS policy: Cross origin requests are only supported for protocol schemes: http, data, chrome, chrome-extension, chrome-untrusted, https.
```

You can get DHT to work, but the connection on the inviting end doesn't get saved as a DHT-type connection and does not reconnect after a refresh.

arj03:
The updatePartialState error is strange. That is only called on `SSB.net.feedReplication` like in profile, I don't know why SSB.net would be undefined. Maybe while it was loading or something.

About that DHT error, you might need to host the files a bit differently now. The browsers have gotten more restrictive so a simple npx http-server doesn't work anymore. I'm using nginx and this config:

```
server {
        listen 8080;
        
        root /home/arj/dev/ssb-browser-demo/dist/;

        index index.html;

        location / {
                # First attempt to serve request as file, then
                # as directory, then fall back to displaying a 404.
                try_files $uri $uri/ =404;

                add_header Cross-Origin-Opener-Policy same-origin;
                add_header Cross-Origin-Embedder-Policy require-corp;
                add_header Cross-Origin-Resource-Policy same-origin;
                # CORS
                #add_header Access-Control-Allow-Origin *;
        }
}
```

KyleMaas:
Those errors popped up after it had been running for probably 2 minutes.  Other functionality seemed to be working and it was talking with the pub, so I don't know either why `SSB.net` would be undefined.

CORS policy doesn't explain why ssb-browser-demo would be trying to issue an `XMLHttpRequest` for the `dht:` address.  Something is causing it to try to ask for that directly instead of running it over the DHT connection.

arj03:
I have just testing DHT between a firefox instance and a chrome, and it is working in this branch for me.

PR #299, opened 2021-06-29
Update ToastUI-Vue editor 3.0.0

### LINKED ISSUE #298: Update vue editor
Hey @KyleMaas 

I noticed that the Vue editor has a new major release that is quite a bit smaller. I was wondering if that would be something you would be up for upgrading?

KyleMaas:
This partially works.  Problems so far include:

* ~Clicking a suggestion in a popup suggestion box (like for channels or people) does not work on the first try.  Subsequent suggestion clicks do seem to work.~
* ~Adding images does not work properly.~

Fixes #298

### DIFF
--- a/browser.html
+++ b/browser.html
@@ -21,7 +21,6 @@
     <link rel="stylesheet" type="text/css" href="css/modal.css">
 
     <link rel="stylesheet" type="text/css" href="node_modules/vue-select/dist/vue-select.css">
-    <link rel="stylesheet" type="text/css" href="node_modules/codemirror/lib/codemirror.css">
     <link rel="stylesheet" type="text/css" href="node_modules/@toast-ui/editor/dist/toastui-editor.css">
     <link rel="stylesheet" type="text/css" href="node_modules/vue-simple-context-menu/dist/vue-simple-context-menu.css">
   </head>
--- a/css/main.css
+++ b/css/main.css
@@ -490,3 +490,8 @@ textarea.source
   height: 20em;
   white-space: nowrap;
 }
+
+.markdown-editor
+{
+  background-color: #fff;
+}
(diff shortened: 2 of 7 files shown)


arj03:
Great first pass there. I tried checking out this branch and building, it's 150kb smaller. And that is not including the css file one doesn't need to include.

KyleMaas:
That pretty much works.  Found an interesting bug I need to file, but it's unrelated to this.  Think I'm going to take this out of Draft status now.  Make sure you test this well before merging, though - the underlying editor is considerably different, so I'm not confident I've handled all edge cases.

KyleMaas:
Inlined size before: 7,831,765 bytes

Inlined size after: 12,193,483 bytes

Something weird going on...

KyleMaas:
Not sure what's going on with inlining.  I'm not seeing much (if any) difference between release build sizes between the master branch and this one.  So if you're good with how this works, go for it, and I'll try and troubleshoot inlining later.

arj03:
Btw one of the audit errors is from the inline tool.

arj03:
I'm testing this now, seems to be working, will fix #300 before merging.

arj03:
This looks really good. Thanks for taken on this. The js reduction is quite substantial. The css is a bit bigger, but not anything close to the increase you are seeing in the inline tool.

KyleMaas:
For whatever reason, this merge is messing up pulling master.

arj03:
What, that's strange. Works fine here. Sure you are on master? :)

KyleMaas:
Yep.  Don't know what's going on, but something's goofed up badly.  Wiping and re-cloning to see if that works.

KyleMaas:
Not sure what happened but wiping and reloading seems to have fixed it.

PR #306, opened 2021-07-06
Fix issue where profile description could cause an error

### LINKED ISSUE #305: Error if profile doesn't have a description
```
Uncaught TypeError: Cannot read property 'matchAll' of undefined
    at showProfile (bundle-ui.js:148339)
    at tryToLoadProfile (bundle-ui.js:148377)
    at bundle-ui.js:148386
```

Stems from here:

https://github.com/arj03/ssb-browser-demo/blob/a1274fdf25a8cb3557d20779578d4fa5be663290/ui/profile.js#L657

KyleMaas:
So...in trying to fix #305, it turns out that we were already checking if the description was undefined, and `self` was already set to `this`.  There were no code paths I could find which would result in `this.descriptionText` to actually be `undefined`.  But it was possible for the error to show up if a profile's description was something other than a string where `matchAll` was not defined on it.  So under the premise that that's probably actually what was causing that error, this changes the `this` reference to `self` for consistency and adds a check to ensure that the description is actually a string before trying to run the regular expression on it.

Fixes #305

### DIFF
--- a/ui/profile.js
+++ b/ui/profile.js
@@ -647,14 +647,14 @@ module.exports = function () {
         if (profile.name)
           self.name = profile.name
           
-        if (profile.description) {
+        if (profile.description && typeof self.descriptionText == "string") {
           self.descriptionText = profile.description
             
           if (self.feedId == SSB.net.id) {
             // Editing self.
             // Check for images.  If there are any, cache them.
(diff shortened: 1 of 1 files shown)


arj03:
That looks great. Thanks :)

PR #309, opened 2021-07-06
Don't try to display information we don't have when displaying an OOO message

### LINKED ISSUE #295: 0 seconds ago
If a message is outside the follow graph, the message shows as being posted "0 seconds ago".  Clicking "Get msg" does not change this - it still says 0 seconds.

KyleMaas:
For OOO messages, we don't have author or timestamp information.  A user also should not be able to react to messages when we have no information about the message other than its ID and the fact that it exists somewhere.

Fixes #295


arj03:
Great!

PR #310, opened 2021-07-06
Fix missing "Mentions:" translations

### LINKED ISSUE #308: common.Mentions
Dead simple to fix, but I don't want to forget it.

ssb-msg seems to be trying to pull a translation string for `common.Mentions` which is not being found, so it just shows up on screen as "common.Mentions".  Needs to be fixed.

KyleMaas:
Quick note: I included the colon on the end in the translation itself, since Japanese uses a different (fixed-width) character for colons than English and Portuguese.

Fixes #308

### DIFF
--- a/messages.json
+++ b/messages.json
@@ -40,6 +40,7 @@
       "forkedreplies": "replies in a forked thread",
       "getMsg": "get msg",
       "Add": "Add",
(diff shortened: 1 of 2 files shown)


arj03:
Thanks

PR #311, opened 2021-07-06
When we can't fetch a thread, display an error

### LINKED ISSUE #264: Viewing a thread seems to hang until connected
It appears (and I haven't debugged this, so I don't know why this would be) that if you're not connected to another peer with data and you try to view a thread, the Thread page will wait (with no indication it's doing anything) until you're connected.  Then it loads.  Only thing I can think of is that maybe it's hanging at the stage where it tries to get a tangle and rather than give up after a bit and query the database directly it just stops.  But that's just a hypothesis - really needs proper debugging.

KyleMaas:
Previously, if you were offline and tried to load a thread with a root you didn't have, it would just freeze.

This makes it so that it displays a useful message to the user and gives them a button to try fetching it again, in case the connection being down was only a temporary thing.

Fixes #264

### DIFF
--- a/messages.json
+++ b/messages.json
@@ -218,6 +218,7 @@
       "loading": "Loading messages...please wait.",
       "postReply": "Post reply",
       "blankFieldError": "Please provide a message to publish.",
+      "cantLoadOOOError": "We don't have this message in the database and are not connected to any peers who might have it.",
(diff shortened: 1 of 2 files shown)


arj03:
Great, simple fix. Just the way I like 'em :)

PR #312, opened 2021-07-06
Check message ID and profile ID validity before trying to navigate

### LINKED ISSUE #270: Validate message ID before trying to navigate to it
Just tried to navigate to a thread by ID, but there was a typo in the ID I copied which made the message ID invalid.  Had me puzzled as to why I couldn't see it.  Would be nice to warn the user that they're trying an invalid message ID and it will never show up.

KyleMaas:
Validate message and feed IDs before trying to blindly navigate to them.

Fixes #270

### DIFF
--- a/messages.json
+++ b/messages.json
@@ -41,6 +41,7 @@
       "getMsg": "get msg",
       "Add": "Add",
       "mentions": "Mentions:",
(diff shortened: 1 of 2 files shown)


arj03:
Looks good, thanks

PR #316, opened 2021-07-08
Add mnemonic key recovery to onboarding

### LINKED ISSUE #255: Load mnemonic key from onboarding screen
Probably would be a good option to allow for onboarding to take in a mnemonic key to restore an account rather than having users set up a new profile and then go to the profile page to load in a key.  I've done it a bunch of times, but thinking about it, it's a non-obvious UI flow.

Implementation-wise, it's not the easiest thing to do, since IIRC the new key is actually generated before the onboarding screen is shown.  So we'd need to generate a new key, show the onboarding screen, replace the key, then reload the page.

Not difficult.  Wouldn't be a bad first project for somebody.

KyleMaas:
Allow the user to choose to onboard as a new user or go straight to loading from a mnemonic key.

Fixes #255

### DIFF
--- a/browser.html
+++ b/browser.html
@@ -4,6 +4,13 @@
     <meta content="text/html;charset=utf-8" http-equiv="Content-Type">
     <meta name="viewport" content="width=device-width,initial-scale=1,maximum-scale=1,user-scalable=0,viewport-fit=cover">
     <link rel="manifest" href="manifest.json">
+
+    <!-- Include these first so they can be overridden. -->
+    <link rel="stylesheet" type="text/css" href="node_modules/vue-select/dist/vue-select.css">
+    <link rel="stylesheet" type="text/css" href="node_modules/@toast-ui/editor/dist/toastui-editor.css">
+    <link rel="stylesheet" type="text/css" href="node_modules/vue-simple-context-menu/dist/vue-simple-context-menu.css">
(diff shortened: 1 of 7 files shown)


KyleMaas:
I'll note that I filed this bug due to a quirk I had to work around to get this to work:

https://github.com/egoist/vue-slim-tabs/issues/141

KyleMaas:
It works properly now.  But it did take some trial and error to work around that bug.  At least the bug was consistent so I could.  :)

arj03:
This is great, can't count how many times I have needed something like this :) I'll let this simmer a day or two and see if upstream responds to your issue.

KyleMaas:
Thanks!  I already worked around that issue, so it shouldn't affect this.  It was just a long road finding what exactly triggered it.  In this case, it happened if I put anything after the `<textarea>` on the second tab.  If I did, then you could switch to that tab but couldn't switch back.  So I just moved the WARNING and NOTE messages outside of the tab area, and now everything's hunky dory.  But it is a bug to be aware of if you ever try to add anything to that tab.

(I did look at other tab libraries.  But this one seemed to be about the simplest, most stable, and smallest I could find.  So working around the bug seemed to be worth it.)

arj03:
Great work! I just tested this and work perfectly.

KyleMaas:
Awesome.  Thanks!

PR #318, opened 2021-07-09
Fix issue where posting a new message popped up the new message indicators

### LINKED ISSUE #302: Posting a message triggers the new message indicator to show up
If you post a new message, odds are you've seen it, so we probably shouldn't be displaying the new message indicator for self-authored messages.

KyleMaas:
When you post a new message, it's pretty obvious there will be a new message.  There's no need to show the new message indicators and cause the user to click them and refresh to see what they are, only to find out it was just their own message.

Fixes #302

### DIFF
--- a/ui/new-private-messages.js
+++ b/ui/new-private-messages.js
@@ -35,11 +35,16 @@ module.exports = function (state) {
           self.registeredSSBChange = true
         }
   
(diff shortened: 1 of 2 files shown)


arj03:
Yes, makes sense. Thanks :)

PR #325, opened 2021-11-08
Port to vanilla branch of ssb-browser-core

KyleMaas:
* Patch dependencies so this can actually be built and tested.
* Update to new SSB singleton setup calls.


KyleMaas:
"Connection status" always shows "Waiting for connection" and 0% complete now, despite it actually continuing to sync.

KyleMaas:
That was a very large API change.

arj03:
> That was a very large API change.

Yes :) But for the better I think.

PR #326, opened 2021-11-08
Add fix and diagnostic message for goofy title crash

### LINKED ISSUE #324: Message title crash if message text doesn't have lastIndexOf()
Ran into this today:

```
Uncaught (in promise) TypeError: msg.content.text.lastIndexOf is not a function
    at Object.exports.getMessageTitle (bundle-ui.js:160711)
    at bundle-ui.js:164267
    at Array.map (<anonymous>)
    at bundle-ui.js:164265
    at bundle-ui.js:24081
    at bundle-ui.js:23977
    at bundle-ui.js:25497
    at bundle-ui.js:25430
    at bundle-ui.js:24278
    at bundle-ui.js:25430
```

Comes from here:

https://github.com/arj03/ssb-browser-demo/blob/584a6bfa3e11f85821d7d2dc2cf7444416b3e553/ui/helpers.js#L43

Execution can only get to that point if `msg.content.text` is not undefined, so evidently there are cases where it can be something other than a string.  Have not figured out which message causes this crash, but we should add an extra validity check.

KyleMaas:
Fixes #324

### DIFF
--- a/ui/helpers.js
+++ b/ui/helpers.js
@@ -36,6 +36,11 @@ exports.getMessageTitle = function(msgId, msg) {
     if (msg.content.title)
       return msg.content.title
     else if (msg.content.text) {
+      if (!msg.content.text.lastIndexOf) {
+        console.log("Encountered message with text not a string:")
+        console.log(msg)
+        return msgId
+      }
       const maxLength = 40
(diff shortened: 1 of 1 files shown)


arj03:
Thanks

